PR #21, opened 2023-05-30
fastsimrust is now a submodule of fastsim

calbaker:
this solves various module not found issues
and is consistent with altrios.
Made it so that tests don't need any dev dependencies.


calbaker:
@nreinicke, could I ask you to take a look at the failing tests for this PR?  I'm unable to reproduce the error on my end and very baffled by the discrepancy.

nreinicke:
@calbaker - What's your definition of done here? I was hitting the same error locally using pytest and so I just switched over to using the unittest library and everything is working. Is there a specific reason you need to use pytest to run the test suite?

Of note: I could get pytest to work when installing the rust extension in a develop mode (either `pip install -e .` or `maturin develop`) and so you should still be able to use pytest locally for development

nreinicke:
@calbaker - FYI - looks like using a dev install with pytest works fine on the server.

kylecarow (approved):
Gave this a spin in a fresh Conda environment and everything seems to work beautifully. Nice!!

PR #25, opened 2023-06-12
209 pyo3 simdrivelabel

Romia87:


rust/fastsim-core/src/vehicle.rs
```diff
@@ -49,45 +51,26 @@ lazy_static! {
         self.mc_peak_eff()
     }
 
-    // TODO: refactor this to have a non-py and `_py` version
```
calbaker, line 52: @Romia87 , I made some edits to your changes and the code is correct now.  I had actually left that `TODO` note for myself months (maybe even years ???) ago and forgotten about it so it's good you got this to happen!

rust/fastsim-core/src/vehicle.rs
```diff
@@ -49,45 +51,26 @@ lazy_static! {
...
-    pub fn set_mc_peak_eff(&mut self, new_peak: f64) {
-        let mc_max_eff = ndarrmax(&self.mc_eff_array);
-        self.mc_eff_array *= new_peak / mc_max_eff;
-        let mc_max_full_eff = arrmax(&self.mc_full_eff_array);
-        self.mc_full_eff_array = self
-            .mc_full_eff_array
-            .iter()
-            .map(|e: &f64| -> f64 { e * (new_peak / mc_max_full_eff) })
-            .collect();
+    #[setter("mc_peak_eff")]
```
calbaker, line 54: `#[setter("mc_peak_eff")]` makes it so that can run `veh.mc_peak_eff = 0.44` in Python.  `set_mc_peak_eff_py` then calls `set_mc_peak_eff`, the pure Rust version.

rust/fastsim-core/src/vehicle.rs
```diff
@@ -37,6 +37,8 @@ lazy_static! {
 #[add_pyo3_api(
```
calbaker, line 37: @Romia87 , any arguments (e.g. function definitions) that get provided to `add_pyo3_api` _do not get compiled_ unless the [`fastsim-py`](https://github.com/NREL/fastsim/tree/fastsim-2/rust/fastsim-py) crate is being compiled because of [this feature](https://github.com/NREL/fastsim/blob/26edd86c100a03fc31586da2e93f5cfbd4fa4d63/rust/fastsim-core/proc-macros/src/add_pyo3_api.rs#L267).  As such, we want the pure rust versions of functions to live elsewhere.  I'll drop a note on that next.

rust/fastsim-core/src/vehicle.rs
```diff
@@ -578,6 +562,32 @@ impl RustVehicle {
         arrmax(&self.fc_eff_array)
     }
 
+    pub fn set_mc_peak_eff(&mut self, new_peak: f64) {
```
calbaker, line 565: @Romia87 , this where the pure rust functions live, and in this location, they get compiled every time no matter what features are enabled or disabled.  

We created the `add_pyo3_api` attribute-style procedural macro because it allows us to automate a lot of the `pyo3` features for exposing struct fields to python.  It also allow use to pass the [pymethods](https://docs.rs/pyo3/latest/pyo3/attr.pymethods.html) block as an argument so that all the `pymethods` definitions can live in a single `impl` block, which is a requirement of pyo3.  It took months and lots of tedious troubleshooting to get this worked out.

rust/fastsim-core/src/simdrivelabel.rs
```diff
@@ -704,7 +709,7 @@ pub fn get_label_fe_phev(
         };
     }
 
-    Ok(phev_calcs)
+    return Ok(phev_calcs);
```
calbaker: The idiomatic way to return in rust is `Ok(phev_calcs)` so please change `return Ok(phev_calcs);` back to `Ok(phev_calcs)` since it will otherwise give warnings whenever we run `cargo clippy`.

calbaker:
@Romia87, [this pymethods attribute](https://github.com/NREL/fastsim/blob/98964069510f291978cd95d97cbb285c4ab698b2/rust/fastsim-core/src/simdrivelabel.rs#L51) and [this pymethods attribute](https://github.com/NREL/fastsim/blob/98964069510f291978cd95d97cbb285c4ab698b2/rust/fastsim-core/src/simdrivelabel.rs#L270) are redundant, unnecessary, and not allowed because that's taken care of in the [add_pyo3_api](https://github.com/NREL/fastsim/blob/98964069510f291978cd95d97cbb285c4ab698b2/rust/fastsim-core/proc-macros/src/add_pyo3_api.rs#L266) procedural macro.  

Also, none of [these setters and getters](https://github.com/NREL/fastsim/blob/98964069510f291978cd95d97cbb285c4ab698b2/rust/fastsim-core/src/simdrivelabel.rs#L52), [thees getters and setters](https://github.com/NREL/fastsim/blob/98964069510f291978cd95d97cbb285c4ab698b2/rust/fastsim-core/src/simdrivelabel.rs#L271), or [these getters and setters](https://github.com/NREL/fastsim/blob/98964069510f291978cd95d97cbb285c4ab698b2/rust/fastsim-core/src/simdrivelabel.rs#L356) are necessary because inside of Rust, the fields of PHEVCycleCalc are all `pub` so we can do stuff like:
```rust
let mut phev_cycle_calc = PHEVCycleCalc::default();
phev_cycle_calc.cd_fs_gal = 2.2;
println!("{}", phev_cycle_calc.cd_fs_gal);
```
without needing getters and setters. 

Additionally, the `add_pyo3_api` procedural macro automatically generates the setters and getters for the python side [here](https://github.com/NREL/fastsim/blob/98964069510f291978cd95d97cbb285c4ab698b2/rust/fastsim-core/proc-macros/src/add_pyo3_api.rs#L87) automatically.

PR #26, opened 2023-06-13
copied better `SerdeAPI` from altrios

calbaker:


rust/fastsim-core/src/traits.rs
```diff
@@ -3,9 +3,23 @@ use std::collections::HashMap;
 
 pub trait SerdeAPI: Serialize + for<'a> Deserialize<'a> {
     /// runs any initialization steps that might be needed
-    fn init(&mut self) {}
+    fn init(&mut self) -> anyhow::Result<()> {
+        Ok(())
+    }
 
-    fn to_file(&self, filename: &str) -> anyhow::Result<()> {
```
kylecarow, line 8: Neat! Didn't know anyhow had its own `Result` built in. That's convenient.

kylecarow (approved):
Looks good!

PR #30, opened 2023-06-16
Vehicle Import Tool

michael-okeefe:
Migrated from https://github.nrel.gov/MBAP/fastsim/pull/237#issuecomment-50921
Migrated also from https://github.nrel.gov/MBAP/fastsim/pull/264

(Apologies for the duplication)

Solves https://github.nrel.gov/MBAP/fastsim/issues/189

Remaining work:

- [x] Clean up code (or annotate with `#[allow(dead_code)]`)
- [x] Add zipped 22-tstcar.csv and enable processing of zip in Python and/or Rust
- [x] Create Github data repository to hold vehicle import data for FASTSim (**need assistance to complete this item**)


michael-okeefe:
As an update, I used the `#[allow(dead_code)]` annotation over some of Virinchi's original functions that were throwing dead code warnings. This will disable those warnings but allow us to determine if that code is, in fact, needed or not. I would recommend pushing off dealing with zip for 22-tstcar.csv until after the import milestone has been completed.

(CC @calbaker )

michael-okeefe:
@kylecarow Looks great, Kyle. My only comment was I noticed you had commented out the glider_kg calculation on line 1031 of vehicle_utils.rs with a TODO -- is that anything we need to address prior to merging? Otherwise, I think it is ready.

kylecarow:
@michael-okeefe 

I'm still a bit hazy on how exactly we match between FE.gov and EPA data, it seems like we narrow down by year/make/model, but there is also some 'scoring' in the code. Would you be able to describe the matching? Or, perhaps better, could you document it in the docstring of `match_epatest_with_fegov_v2`?

kylecarow:
> @kylecarow Looks great, Kyle. My only comment was I noticed you had commented out the glider_kg calculation on line 1031 of vehicle_utils.rs with a TODO -- is that anything we need to address prior to merging? Otherwise, I think it is ready.

I noticed that the back-calculation of mass didn't properly add up to the EPA test mass, so I added the EPA test mass as an override and ignored the back-calculated `glider_kg`; It would be nice to have this fixed eventually so that component masses are 'correct', but I think ensuring the override mass is the same as the EPA test mass is the way to go anyway. Basically I don't think it should block merging this PR.

michael-okeefe:
> @michael-okeefe
> 
> I'm still a bit hazy on how exactly we match between FE.gov and EPA data, it seems like we narrow down by year/make/model, but there is also some 'scoring' in the code. Would you be able to describe the matching? Or, perhaps better, could you document it in the docstring of `match_epatest_with_fegov_v2`?

@kylecarow 
I just pushed up a docstring with the explanation. Please let me know if you have any questions.

kylecarow:
> @kylecarow I just pushed up a docstring with the explanation. Please let me know if you have any questions.

Awesome, thanks Michael!

kylecarow (approved):

PR #38, opened 2023-08-02
F2 doc fields

calbaker:
Added proc macro to create `*_doc: Option<String>` for most vehicle fields that can be skipped with `#[doc_field(skip_doc)]` attribute.  

TODO:
- [x] make sure `*_doc` fields are accessible from python
- [x] publish `fastim-proc-macros` with new version
- [x] fix Cargo.toml path to not be local
- [x] improve documentation in fastsim-proc-macros/lib.rs


calbaker:
@kylecarow , this isn't ready to merge, but it's a decent idea for you to play with it.

kylecarow:
After discussion today, it seems we are in agreement that a separate doc struct is the way to go for this

PR #41, opened 2023-08-04
Add defaults to `regen_a` and `regen_b` for serde

kylecarow:
I've run the tests as well as messed with parameters to confirm that this does what I expect.


calbaker (approved):
Nice work!

PR #43, opened 2023-08-10
Dummy PR: changed to `drag_coef: VehicleField<f64>`

calbaker:
some tests fail but most pass
despite having `std::ops::Mul` implemented, requires unwieldy use of `clone()` method to satisfy borrow checker

### DIFF
--- a/rust/fastsim-core/src/simdrive/cyc_mods.rs
+++ b/rust/fastsim-core/src/simdrive/cyc_mods.rs
@@ -332,8 +332,9 @@ impl RustSimDrive {
             };
             let g = self.props.a_grav_mps2;
             let m = self.veh.veh_kg;
-            let rho_cdfa =
-                self.props.air_density_kg_per_m3 * self.veh.drag_coef * self.veh.frontal_area_m2;
+            let rho_cdfa = self.veh.drag_coef.clone()
+                * self.props.air_density_kg_per_m3
+                * self.veh.frontal_area_m2;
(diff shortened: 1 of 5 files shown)


rust/fastsim-core/src/simdrive/cyc_mods.rs
```diff
@@ -332,8 +332,9 @@ impl RustSimDrive {
             };
             let g = self.props.a_grav_mps2;
             let m = self.veh.veh_kg;
-            let rho_cdfa =
-                self.props.air_density_kg_per_m3 * self.veh.drag_coef * self.veh.frontal_area_m2;
+            let rho_cdfa = self.veh.drag_coef.clone()
```
calbaker, line 335: @kylecarow , `VehiclField<f64>` can be multiplied by f64 but not the other order.  Because types that are used in math operations normally implement the `std::markers::Copy` trait and `VehicleField<f64>` cannot implement that trait due to fields that can't do it, we need to use `.clone()` a lot, which in my mind negates the benefit of the `std::ops` implementations.

rust/fastsim-core/src/vehicle_utils.rs
```diff
@@ -180,6 +183,130 @@ impl CostFunction for GetError<'_> {
     }
 }
 
+macro_rules! array_and_scalar_ops {
```
calbaker, line 186: @kylecarow , this is where I've implemented the `std::ops` traits.

PR #46, opened 2023-08-17
F2 aux override speedup

calbaker:
- creates new `RustSimDrive` [field]:(https://github.com/NREL/fastsim/blob/c0f12f9d104e22297045ec0691975c96f873a4a2/rust/fastsim-core/src/simdrive.rs#L495)
```
#[api(skip_get, skip_set)]
#[serde(skip)]
aux_in_kw_override: Option<Vec<f64>>,
```
that allows for a persistent array to eliminate the need for repeated cloning.  
- uses matching to cheaply figure out if `aux_in_kw` needs override:
https://github.com/NREL/fastsim/blob/c0f12f9d104e22297045ec0691975c96f873a4a2/rust/fastsim-core/src/simdrive/simdrive_impl.rs#L529

### DIFF
--- a/build_and_test.sh
+++ b/build_and_test.sh
@@ -1,4 +1,5 @@
-(cd rust/fastsim-core/ && cargo test) && \
+# build and test with local version of `fastsim-proc-macros`
+(cd rust/fastsim-core/ && cargo test --features dev-proc-macros) && \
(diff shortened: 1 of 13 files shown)


calbaker:
@kylecarow , this is ready for review

kylecarow (approved):

PR #55, opened 2023-10-04
adding demo files to build_and_test.sh run

robinsteuteville:
I added a run_all.py file and added that to the build_and_test.sh file. As long as I x out of the graphs/tables that are made as the sh file runs the demo files, the sh file seems to properly run the files (if I don't x out of the graph/table, it will stop running until I do so). One of the files is throwing errors when I try to run them (see below), so I will now move on to troubleshooting those.


robinsteuteville:
For the accel_demo.py file, running the file gives these warnings (not sure if the warnings are ok or not):
image

robinsteuteville:
for cav_demo.py, it ran successfully (RAN_SUCCESSFULLY = True), but had the following warning: 
image

robinsteuteville:
demo_abc_drag_coef_conv.py ran with the following warnings: 
image

robinsteuteville:
demo.py ran with the following warnings:
image
image
image

robinsteuteville:
fusion_thermal_demo.py was not able to run due to the following error:
image

robinsteuteville:
mp_parallel_demo.py ran with the following warning:
2023-10-04 16:31:26.248007
Running with 4 pool processes.
<ipython-input-1-10feee12782c>:42: UserWarning: Unknown arguments: ['--ip=127.0.0.1', '--stdin=9013', '--control=9011', '--hb=9010', '--Session.signature_scheme="hmac-sha256"', '--Session.key=b"b751cf54-1710-4444-92d2-6af34db5352e"', '--shell=9012', '--transport="tcp"', '--iopub=9014', '--f=c:\\Users\\rsteutev\\AppData\\Roaming\\jupyter\\runtime\\kernel-v2-1832nG2N0vD4uUgT.json']
  warnings.warn(f"Unknown arguments: {unknown}")
Elapsed time: 36.17

robinsteuteville:
time_dilation_demo.py ran with the following warnings: 
image

robinsteuteville:
test_demo.py failed to run with the following error:
image

robinsteuteville:
test_cav_demo.py failed to run with the following error:
image

robinsteuteville:
The following files run with warnings: 
accel_demo.py
cav_demo.py
demo_abc_drag_coef_conv.py
demo.py
mp_parallel_demo.py
time_dilation_demo.py
The following files ran without any warnings or errors:
demo_eu_vehicle_wltp.py
stop_start_demo.py
timing_demo.py

robinsteuteville:
@calbaker I think I successfully created a file run_all.py which finds all files in the demos folder which have 'demo' in the title and end in .py and runs them. I added this file to the build_and_test.sh so that it runs whenever the build_and_test.sh file is run. A couple things: right now, each of the demo files runs fully, which leads to many lines of output, making it hard to tell which files have run and, if the overall build_and_test.sh doesn't run, which file caused the error. Would it be better to figure out a way to run each file which only returned whether or not the file successfully ran, rather than all the demo file's outputs? Not sure how to do this off the top of my head but happy to work on this if you think it would be helpful. Also, is there anything I should do in order to make it so that the demo folder doesn't get downloaded with the program when fastsim is installed with pip, and instead is downloaded separately like in altrios?

python/fastsim/__init__.py
```diff
@@ -33,6 +33,39 @@ def package_root() -> Path:
 
 __version__ = get_distribution("fastsim").version
 
+#download demo files
```
robinsteuteville: @calbaker this is what I have so far, but I'm somewhat stumped, any ideas?

robinsteuteville:
The latest commit implements altrios-style demos testing in the build_and_test.sh file. The sh file now runs all of the demos files, without showing the plots (a SHOW_PLOTS variable is used to determine whether or not the plots are shown, with the variable being false for the sh tests). Also partially implemented is a function in the utilities file that downloads the demos files (so that users can download and modify demos files). @calbaker how do I add the demos check to the github tests? Not sure how to get started with that. Also, there are two files in the tests folder -- test_cav_demo.py and test_demo.py that I believe are now redundant and can be deleted. It seems to me like these to files exist to test the cav_demo.py and demo.py files, which are now being tested along with the other files. Does this seem correct?

python/fastsim/utilities.py
```diff
@@ -394,3 +397,54 @@ def calculate_tire_radius(tire_code: str, units: str = "m"):
...
     return radius
+
+def show_plots() -> bool:
+    """
+    Returns true if plots should be displayed
+    """
+    return os.environ.get("SHOW_PLOTS", "true").lower() == "true"     
+
+#do I need a different default path?
+def download_demo_files(demo_path: Path=Path("demos")):
```
robinsteuteville: @calbaker any suggestions for how to go about testing/troubleshooting this to see if it works?

robinsteuteville:
Most recent commit adding "requests" python library to pyproject.toml made it so all the github checks passed (before, none of them did).

python/fastsim/utilities.py
calbaker: @robinsteuteville , did you find my ALTRIOS test for this?  Also, since you'll be copying rather than downloading, this is resolved right?

robinsteuteville: @calbaker yes to both those questions

robinsteuteville:
Not sure why, but my fastsim code stopped being able to compile today, with this error:
`#![feature]` may not be used on the stable release channel
This was referring to line of code 238 in lib.rs:
#![cfg_attr(error_generic_member_access, feature(error_generic_member_access))]
Not sure why this error is coming up now, since there have been no changes on this branch to the lib.rs file. This means I can't test/troubleshoot the copy_demo_files function -- or at least, I haven't figure out a way around this in order to do so. I tried commenting out this one line of code, but it just added more errors. Any idea what is going wrong @calbaker @kylecarow ?

calbaker:
@robinsteuteville , I think you could actually probably figure this one out based on spending ~20 minutes doing initial troubleshooting

.github/workflows/tests.yaml
```diff
@@ -36,4 +38,5 @@ jobs:
 
       - name: Python unit tests 
         run: |
-          pip install -e ".[dev]" && pytest -v python/fastsim/tests/
+          pip install -e ".[dev]" && pytest -v python/fastsim/tests/ 
+          pip install -e ".[dev]" && pytest -v python/fastsim/demos/
```
calbaker: I think this line can be reduced to just  
`pytest -v python/fastsim/demos/` 
@robinsteuteville , could you try that out?

python/fastsim/__init__.py
```diff
@@ -29,11 +29,10 @@ def package_root() -> Path:
 logger = logging.getLogger(__name__)
 
 from pkg_resources import get_distribution
-
 __version__ = get_distribution("fastsim").version
 
-__doc__ += "\nhttps://pypi.org/project/fastsim/"
-__doc__ += "\nhttps://www.nrel.gov/transportation/fastsim.html"
+# __doc__ += "\nhttps://pypi.org/project/fastsim/"
```
calbaker: @robinsteuteville , these were actually appending these lines to the module level doc string, which is something we'd definitely like to retain.  Please uncomment.

robinsteuteville: @calbaker I believe this change happened when I merged fastsim-2 main branch into my branch. Are the deletions of lines 35 and 36 ok (also from the merge)?

python/fastsim/__init__.py
```diff
@@ -29,11 +29,10 @@ def package_root() -> Path:
...
+# __doc__ += "\nhttps://www.nrel.gov/transportation/fastsim.html"
```
kylecarow:
```suggestion
__doc__ += "\nhttps://www.nrel.gov/transportation/fastsim.html"
```

python/fastsim/__init__.py
kylecarow:
```suggestion
__doc__ += "\nhttps://pypi.org/project/fastsim/"
```

.github/workflows/tests.yaml
kylecarow:
```suggestion
          pytest -v python/fastsim/demos/
```

python/fastsim/__init__.py
calbaker: @robinsteuteville , is this thread resolved?  If yes, mark the 
image
button below.

calbaker:
@robinsteuteville , is this one ready to go?

robinsteuteville:
> @robinsteuteville , is this one ready to go?

@calbaker I believe so. I fixed the issues that you noted earlier -- perhaps check to make sure they seem good to you, and if so, it should be ready to merge.

calbaker (approved):

PR #56, opened 2023-10-10
Replace Option::unwrap() with Option::ok_or_else() with descriptive message

kylecarow:
<strike>
The only thing I would want your input on this the verbiage of the new error message:

`File extension not specified`

versus the possible times a `None` is returned by `PathBuf::extension()`:
```
The extension is:
- None, if there is no file name;
- None, if there is no embedded .;
- None, if the file name begins with . and has no other .s within;
- Otherwise, the portion of the file name after the final .
```
</strike>

**edit: I think the updated error message is fine.**

### DIFF
--- a/rust/fastsim-core/src/traits.rs
+++ b/rust/fastsim-core/src/traits.rs
@@ -19,7 +19,7 @@ pub trait SerdeAPI: Serialize + for<'a> Deserialize<'a> {
     /// # Returns:
     ///
     /// A Rust Result
-    fn to_file(&self, filename: &str) -> Result<(), anyhow::Error> {
+    fn to_file(&self, filename: &str) -> anyhow::Result<()> {
         let file = PathBuf::from(filename);
         match file
             .extension()
--- a/rust/fastsim-core/src/traits.rs
+++ b/rust/fastsim-core/src/traits.rs
@@ -25,8 +25,12 @@ pub trait SerdeAPI: Serialize + for<'a> Deserialize<'a> {
             .extension()
             .ok_or_else(|| anyhow!("Unable to parse file extension: {:?}", file))?
             .to_str()
-            .unwrap()
-        {
+            .ok_or_else(|| {
+                anyhow!(
+                    "Unable to convert file extension from `&OsStr` to `&str`: {:?}",
+                    file
+                )
+            })? {
             "json" => serde_json::to_writer(&File::create(file)?, self)?,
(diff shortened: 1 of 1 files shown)


kylecarow:
@calbaker its also possible for OsStr.to_str() to panic at the unwrap, because OsStr can hold non-utf8 data and str cannot

But I don't think it would be worth making a specific error message just for the case someone provides non-utf8 data in the file extension or something

rust/fastsim-core/src/traits.rs
```diff
@@ -19,9 +19,14 @@ pub trait SerdeAPI: Serialize + for<'a> Deserialize<'a> {
     /// # Returns:
     ///
     /// A Rust Result
-    fn to_file(&self, filename: &str) -> Result<(), anyhow::Error> {
+    fn to_file(&self, filename: &str) -> anyhow::Result<()> {
         let file = PathBuf::from(filename);
-        match file.extension().unwrap().to_str().unwrap() {
+        match file
+            .extension()
+            .ok_or_else(|| anyhow!("Unable to parse file extension: {:?}", file))?
```
calbaker, line 26: @kylecarow , I think you could change 
```rust
ok_or_else(|| anyhow!("Unable to parse file extension: {:?}", file))?
```
to something like
```rust
ok_or(|e| anyhow!("{e}\nUnable to parse file extension: {:?}", e, file))?
```
to preserve whatever the native error message is as context.

rust/fastsim-core/src/traits.rs
```diff
@@ -19,9 +19,14 @@ pub trait SerdeAPI: Serialize + for<'a> Deserialize<'a> {
...
+            .to_str()
+            .unwrap()
```
calbaker: @kylecarow , I'd lean toward also getting rid of this unwrap with a separate message since it's not much additional effort.

rust/fastsim-core/src/traits.rs
kylecarow: Hm, I don't know that the native error message really exists. The `unwrap` was on an `Option`, not a `Result`, so the error was just about unwrapping on a `None`

rust/fastsim-core/src/traits.rs
kylecarow: @calbaker done, see latest commit

calbaker (approved):

PR #57, opened 2023-10-13
replace map_err with with_context

kylecarow:
really small PR for ya, couldn't find more instances of this


calbaker (approved):
great work!

PR #58, opened 2023-10-18
Serde revamp and anyhowization

calbaker:


rust/fastsim-cli/src/bin/fastsim-cli.rs
```diff
@@ -290,7 +290,7 @@ pub fn main() {
             traceMissInMph: sdl.0.trace_miss_speed_mph,
             h2AndDiesel: None,
         };
-        println!("{}", res.to_json());
+        println!("{}", res.to_json().unwrap());
```
calbaker: @kylecarow, I'm not certain about this, but maybe it'd be better to have the containing function return a result and use `?` rather than `.unwrap()`

rust/fastsim-cli/src/bin/fastsim-cli.rs
```diff
@@ -358,7 +358,7 @@ pub fn main() {
             traceMissInMph: sim_drive.trace_miss_speed_mps * MPH_PER_MPS,
             h2AndDiesel: h2_diesel_results,
         };
-        println!("{}", res.to_json());
+        println!("{}", res.to_json().unwrap());
```
calbaker: same comment as above

rust/fastsim-core/src/traits.rs
```diff
@@ -35,62 +35,66 @@ pub trait SerdeAPI: Serialize + for<'a> Deserialize<'a> {
...
+            _ => anyhow::bail!("Unsupported file extension {}", extension),
         };
         file_de.init()?;
         Ok(file_de)
     }
 
-    /// json serialization method.
-    fn to_json(&self) -> String {
-        serde_json::to_string(&self).unwrap()
+    // /// JSON serialization method
```
calbaker: @kylecarow , what's going on with this comment block?

rust/fastsim-core/src/imports.rs
```diff
@@ -1,5 +1,5 @@
-pub(crate) use anyhow::{anyhow, bail, ensure};
```
calbaker: @kylecarow , what was your motivation for this change?

kylecarow: @calbaker I figured if we were using anyhow::Result, anyhow::bail!, anyhow::ensure!, anyhow::anyhow!, etc. we might as well fully qualify where they come from. Just changes how the code looks, really (and the ergonomics of using anyhow stuff).

If you prefer it the other way that's fine too, but we use quite a lot of anyhow macros and stuff

rust/fastsim-core/src/traits.rs
kylecarow: @calbaker oops, that's leftover from my debugging, I'll remove it

rust/fastsim-cli/src/bin/fastsim-cli.rs
kylecarow: I agree that this function should return a Result, just put these unwraps to make the code compile (and there are other unwraps in this function already).

I'll change the function to return `anyhow::Result` and change these unwraps to `?`s as appropriate, as that would fit well in this PR

rust/fastsim-cli/src/bin/fastsim-cli.rs
kylecarow: same comment as above!

rust/fastsim-core/src/imports.rs
kylecarow: as for the bincode import change, that just clarifies where `serialize` and `deserialize` in `traits.rs` come from

rust/fastsim-cli/src/bin/fastsim-cli.rs
kylecarow: done!

kylecarow:
@calbaker check out the serde api changes ive been playing with! This PR kind of ballooned into more than just `anyhow`izing things...

### Improvements:
- Python-side calls to the `to_file` and `from_file` methods now also accept `pathlib.Path`s, in addition to strings
  - example:

    ``` python
    # %%
    import fastsim as fsim
    from pathlib import Path

    # %%
    filepath = "hwfet.csv"
    fsim.fsr.RustCycle.from_file(filepath)
    # returns <fastsimrust.RustCycle at 0x12cf84430>

    # %%
    filepath = Path("hwfet.csv")
    fsim.fsr.RustCycle.from_file(filepath)
    # returns <fastsimrust.RustCycle at 0x12cf84530>
    ```
  - see how it works here:
  https://github.com/NREL/fastsim/blob/d0a15aef11d8468ba276e5cccb7a2130fbafaaa5/rust/fastsim-core/fastsim-proc-macros/src/add_pyo3_api/mod.rs#L108-L121

- Rust-side calls to the same methods now accept Rust `Path` objects as well as strings, because I made the argument generic:
https://github.com/NREL/fastsim/blob/d0a15aef11d8468ba276e5cccb7a2130fbafaaa5/rust/fastsim-core/src/traits.rs#L22-L23
https://github.com/NREL/fastsim/blob/d0a15aef11d8468ba276e5cccb7a2130fbafaaa5/rust/fastsim-core/src/traits.rs#L54-L55

# But wait, there's more!

### New methods (not yet Python exposed):

- `from_reader`:
  https://github.com/NREL/fastsim/blob/d0a15aef11d8468ba276e5cccb7a2130fbafaaa5/rust/fastsim-core/src/traits.rs#L98-L104
  - accepts anything that implements `std::io::Read`, for example `std::fs::File`, TCP streams, etc.
  - this method is what `from_file` now uses and offloads extension matching to (see line 73):
    https://github.com/NREL/fastsim/blob/d0a15aef11d8468ba276e5cccb7a2130fbafaaa5/rust/fastsim-core/src/traits.rs#L54-L76
- `from_str`:
  https://github.com/NREL/fastsim/blob/d0a15aef11d8468ba276e5cccb7a2130fbafaaa5/rust/fastsim-core/src/traits.rs#L106-L112
  - reads data from a string (this did end up being the easiest way to load compiled data)
  - since this is part of the trait, it can't also accept a `name` (would have a different function signature), so the `from_csv_str` method still exists which can still do that
    - `RustCycle::from_str` with `format="csv"` can still be used if you don't care about the cycle name, it will just use the `from_csv_str` method with `name = ""`:
    https://github.com/NREL/fastsim/blob/d0a15aef11d8468ba276e5cccb7a2130fbafaaa5/rust/fastsim-core/src/cycle.rs#L658-L667

- `from_resource`:
  https://github.com/NREL/fastsim/blob/d0a15aef11d8468ba276e5cccb7a2130fbafaaa5/rust/fastsim-core/src/traits.rs#L78-L96
  - accepts a path to read from `fastsim-core/resources`
    - for example:
    https://github.com/NREL/fastsim/blob/d0a15aef11d8468ba276e5cccb7a2130fbafaaa5/rust/fastsim-core/src/simdrivelabel.rs#L200-L201
  - uses `from_str` in the backend, which matches based on the filepath extension
  - this should probably be exposed to Python eventually

kylecarow:
@calbaker I would be fine with changing the tests that return a result back to returning `()`, since results in tests aren't doing anything special than allowing `?` in place of `.unwrap()`

rust/fastsim-core/src/imports.rs
calbaker: >  but we use quite a lot of anyhow macros and stuff

That's my very reason for wanting a shorthand.  For imports that are widely used, I tend to prefer that.  For imports that are used only sparsely, I tend to prefer full-path usage.  I'm not super firm on any of this.

kylecarow:
@calbaker I'm happy with this if you consider it complete.

I changed the tests to panic instead of returning results, and the anyhow import is more consistent with the way we've been doing it.

kylecarow:
@calbaker 

I have found a lot of functions that return results, when they don't contain any fallible code. This happens in a lot of the methods called by `solve_step`.

For example:

<details open>
<summary>This code</summary>

```rust
    /// Perform all the calculations to solve 1 time step.
    pub fn solve_step(&mut self, i: usize) -> anyhow::Result<()> {
        self.set_misc_calcs(i)?;
        self.set_comp_lims(i)?;
        self.set_power_calcs(i)?;
        self.set_ach_speed(i)?;
        self.set_hybrid_cont_calcs(i)?;
        self.set_fc_forced_state_rust(i)?;
        self.set_hybrid_cont_decisions(i)?;
        self.set_fc_power(i)?;
        Ok(())
    }

    /// Sets misc. calculations at time step 'i'
    /// Arguments:
    /// ----------
    /// i: index of time step
    pub fn set_misc_calcs(&mut self, i: usize) -> anyhow::Result<()> {
        // if cycle iteration is used, auxInKw must be re-zeroed to trigger the below if statement
        // TODO: this is probably computationally expensive and was probably a workaround for numba
        // figure out a way to not need this
        if self.aux_in_kw.slice(s![i..]).iter().all(|&x| x == 0.0) {
            // if all elements after i-1 are zero, trigger default behavior; otherwise, use override value
            if self.veh.no_elec_aux {
                self.aux_in_kw[i] = self.veh.aux_kw / self.veh.alt_eff;
            } else {
                self.aux_in_kw[i] = self.veh.aux_kw;
            }
        }
        // Is SOC below min threshold?
        self.reached_buff[i] = self.soc[i - 1] >= (self.veh.min_soc + self.veh.perc_high_acc_buf);

        // Does the engine need to be on for low SOC or high acceleration
        self.high_acc_fc_on_tag[i] = self.soc[i - 1] < self.veh.min_soc
            || (self.high_acc_fc_on_tag[i - 1] && !(self.reached_buff[i]));
        self.max_trac_mps[i] =
            self.mps_ach[i - 1] + (self.veh.max_trac_mps2 * self.cyc.dt_s_at_i(i));
        Ok(())
    }
```
</details>

<details open>
<summary>and this code</summary>

```rust
    /// Perform all the calculations to solve 1 time step.
    pub fn solve_step(&mut self, i: usize) -> anyhow::Result<()> {
        self.set_misc_calcs(i);
        self.set_comp_lims(i)?;
        self.set_power_calcs(i)?;
        self.set_ach_speed(i)?;
        self.set_hybrid_cont_calcs(i)?;
        self.set_fc_forced_state_rust(i)?;
        self.set_hybrid_cont_decisions(i)?;
        self.set_fc_power(i)?;
        Ok(())
    }

    /// Sets misc. calculations at time step 'i'
    /// Arguments:
    /// ----------
    /// i: index of time step
    pub fn set_misc_calcs(&mut self, i: usize) {
        // if cycle iteration is used, auxInKw must be re-zeroed to trigger the below if statement
        // TODO: this is probably computationally expensive and was probably a workaround for numba
        // figure out a way to not need this
        if self.aux_in_kw.slice(s![i..]).iter().all(|&x| x == 0.0) {
            // if all elements after i-1 are zero, trigger default behavior; otherwise, use override value
            if self.veh.no_elec_aux {
                self.aux_in_kw[i] = self.veh.aux_kw / self.veh.alt_eff;
            } else {
                self.aux_in_kw[i] = self.veh.aux_kw;
            }
        }
        // Is SOC below min threshold?
        self.reached_buff[i] = self.soc[i - 1] >= (self.veh.min_soc + self.veh.perc_high_acc_buf);

        // Does the engine need to be on for low SOC or high acceleration
        self.high_acc_fc_on_tag[i] = self.soc[i - 1] < self.veh.min_soc
            || (self.high_acc_fc_on_tag[i - 1] && !(self.reached_buff[i]));
        self.max_trac_mps[i] =
            self.mps_ach[i - 1] + (self.veh.max_trac_mps2 * self.cyc.dt_s_at_i(i));
    }
```
</details>

Achieve exactly the same thing, and results aren't giving us any benefit

What's your opinion on removing the `Result` returns and `Ok(())`s in cases like these?

calbaker:
> What's your opinion on removing the Result returns and Ok(())s in cases like these?

@kylecarow, I'd say keep it as is!  On numerous occasions, I've ended up changing functions that do not contain fallible code to contain fallible code, and returning a result that does nothing future proofs for that without imposing any meaningful computational costs, which means that if the function/method body changes, no other api changes are needed.

kylecarow:
@calbaker Sounds reasonable to me. I marked this as ready for review, feel free to merge when tests pass and you're happy with the changes.

kylecarow:
@calbaker didnt you say something about wanting to be able to read/write bincode objects? I could set that up to work with a `.bin` extension (or something) really easily in this branch

kylecarow (approved):

kylecarow:
@calbaker feel free to merge when you're happy with the changes, i added an approving review to let the merge happen

rust/fastsim-core/src/imports.rs
calbaker: @kylecarow , where did we leave off here?

kylecarow: @calbaker That line is now back to `pub(crate) use anyhow::{anyhow, bail, ensure, Context};`

rust/fastsim-core/src/traits.rs
```diff
@@ -13,28 +13,29 @@ pub trait SerdeAPI: Serialize + for<'a> Deserialize<'a> {
...
-            "json" => serde_json::to_writer(&File::create(file)?, self)?,
-            "yaml" => serde_yaml::to_writer(&File::create(file)?, self)?,
-            _ => serde_json::to_writer(&File::create(file)?, self)?,
-        };
+            })?;
+        match extension {
+            "json" => serde_json::to_writer(&File::create(filepath)?, self)?,
+            "yaml" => serde_yaml::to_writer(&File::create(filepath)?, self)?,
+            // TODO: do we want a default behavior for this?
+            _ => serde_json::to_writer(&File::create(filepath)?, self)?,
```
kylecarow: @calbaker can you take a look at this comment?

kylecarow (approved):

kylecarow (approved):

kylecarow:
@calbaker I used `.trim_start_matches('.')` on strings to ignore leading `.`s, which simplifies the match conditions a bit

https://github.com/NREL/fastsim/blob/b8bdfa2e545cb3932a69ef2cb026f169e47708d1/rust/fastsim-core/src/traits.rs#L110-L115

kylecarow:
I added an implementation of `to_file` and `to_str` for `RustCycle`, so we can now write cycle CSV files and strings from Rust 😁

kylecarow:
I'll go ahead and merge this into `fastsim-2`, with the understanding that bincode serde is broken for vehicles and simdrives (as it currently is in `fastsim-2` anyway)

see https://github.nrel.gov/MBAP/fastsim/issues/279 for details

PR #61, opened 2023-10-25
Update set ach speed

calbaker:
Implemented the following in https://github.com/NREL/fastsim/blob/0a4b848acaf2a672d93d8d8dfd495415311757f8/rust/fastsim-core/src/simdrive/simdrive_impl.rs#L953
- more descriptive variable names
- use of closures to reduce redundancy

I did this to make it easier to see how to write this section in `fastsim-3`.

### DIFF
--- a/python/fastsim/demos/__init__.py
+++ b/python/fastsim/demos/__init__.py
@@ -1,4 +1,7 @@
-print(
-    "Module `fastsim.demos` has been imported.  This should only " 
-    + "happen during testing so you might be doing something wrong."
+import logging
+
+logger = logging.getLogger(__name__)
+logger.warning(
+    "Module `fastsim.demos` has been imported. This should only "
+    + "happen during testing, so you might be doing something wrong."
 )
--- a/python/fastsim/demos/__init__.py
+++ b/python/fastsim/demos/__init__.py
@@ -1,0 +1,4 @@
+print(
+    "Module `fastsim.demos` has been imported.  This should only " 
+    + "happen during testing so you might be doing something wrong."
+)
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -1,4 +1,5 @@
 [workspace]
+resolver = "2"
 
 members = [
     "fastsim-cli",                      # command line app
--- a/rust/fastsim-core/src/simdrive/simdrive_impl.rs
+++ b/rust/fastsim-core/src/simdrive/simdrive_impl.rs
@@ -632,20 +632,11 @@ impl RustSimDrive {
             }
         }
         // Is SOC below min threshold?
-        if self.soc[i - 1] < (self.veh.min_soc + self.veh.perc_high_acc_buf) {
-            self.reached_buff[i] = false;
-        } else {
-            self.reached_buff[i] = true;
-        }
+        self.reached_buff[i] = self.soc[i - 1] >= (self.veh.min_soc + self.veh.perc_high_acc_buf);
 
         // Does the engine need to be on for low SOC or high acceleration
-        if self.soc[i - 1] < self.veh.min_soc
-            || (self.high_acc_fc_on_tag[i - 1] && !(self.reached_buff[i]))
-        {
-            self.high_acc_fc_on_tag[i] = true
-        } else {
-            self.high_acc_fc_on_tag[i] = false
-        }
+        self.high_acc_fc_on_tag[i] = self.soc[i - 1] < self.veh.min_soc
+            || (self.high_acc_fc_on_tag[i - 1] && !(self.reached_buff[i]));
         self.max_trac_mps[i] =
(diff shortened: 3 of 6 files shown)


python/fastsim/demos/__init__.py
```diff
@@ -0,0 +1,4 @@
+print(
+    "Module `fastsim.demos` has been imported.  This should only " 
+    + "happen during testing so you might be doing something wrong."
+)
```
kylecarow:
```suggestion
import logging

logger = logging.getLogger(__name__)
logger.warning(
    "Module `fastsim.demos` has been imported. This should only "
    + "happen during testing, so you might be doing something wrong."
)
```

kylecarow: Results in something like this:
```
2023-11-02 14:23:35.117 | __init__.py:4 | WARNING: Module `fastsim.demos` has been imported. This should only happen during testing, so you might be doing something wrong.
```

rust/Cargo.toml
```diff
@@ -1,4 +1,5 @@
 [workspace]
+resolver = "2"
```
kylecarow, line 2: @calbaker do you know what this does? I remember seeing obnoxious messages about it when compiling.

rust/fastsim-core/src/simdrive/simdrive_impl.rs
```diff
@@ -895,13 +886,9 @@ impl RustSimDrive {
...
-            * self.cyc_whl_rad_per_sec[i].powf(2.0)
-            / self.cyc.dt_s_at_i(i)
-            - 0.5
-                * self.veh.wheel_inertia_kg_m2
-                * self.veh.num_wheels
-                * (self.mps_ach[i - 1] / self.veh.wheel_radius_m).powf(2.0)
-                / self.cyc.dt_s_at_i(i))
+            * (self.cyc_whl_rad_per_sec[i].powi(2)
+                - (self.mps_ach[i - 1] / self.veh.wheel_radius_m).powi(2))
+            / self.cyc.dt_s_at_i(i))
```
kylecarow, line 891: @calbaker This change seems like it'd result in a different behavior, could you please elaborate for me?

rust/fastsim-core/src/simdrive/simdrive_impl.rs
```diff
@@ -632,20 +632,11 @@ impl RustSimDrive {
             }
         }
         // Is SOC below min threshold?
-        if self.soc[i - 1] < (self.veh.min_soc + self.veh.perc_high_acc_buf) {
-            self.reached_buff[i] = false;
-        } else {
-            self.reached_buff[i] = true;
-        }
+        self.reached_buff[i] = self.soc[i - 1] >= (self.veh.min_soc + self.veh.perc_high_acc_buf);
```
kylecarow, line 635: Sick!!

rust/fastsim-core/src/simdrive/simdrive_impl.rs
```diff
@@ -1020,56 +1007,74 @@ impl RustSimDrive {
...
-                self.newton_iters[i] = iterate;
+                self.newton_iters[i] = spd_ach_i;
 
-                let _ys = Array::from_vec(ys).map(|x| x.abs());
+                let _ys = Array::from_vec(pwr_errs).map(|x| x.abs());
+                // Question: could we assume `speed_guesses.iter().last()` is the correct solution?
+                // This would make for faster running.
                 self.mps_ach[i] = max(
-                    xs[_ys
+                    speed_guesses[_ys
```
kylecarow, line 1077: I think I'll just trust this is good since the tests pass...

kylecarow (changes requested):
I think the logger stuff is all I care to change, I like the rest of the changes!!

python/fastsim/demos/__init__.py
kylecarow: This also allows this message be silenced by the logging level utilities in `utilities.py`

rust/Cargo.toml
calbaker: @kylecarow , it solves obnoxious messages!

rust/fastsim-core/src/simdrive/simdrive_impl.rs
calbaker: I refactored it from the form `a*b - a * c` to the form `a * (b - c)` to make use of transitive property of multiplication to reduce the number of multiply/divide operations and arguably make the code more readable.

rust/fastsim-core/src/simdrive/simdrive_impl.rs
calbaker: @kylecarow , all thanks to `cargo clippy`

rust/fastsim-core/src/simdrive/simdrive_impl.rs
kylecarow: I hadn't seen lines 887 and 888! All good

kylecarow (approved):

PR #66, opened 2023-11-02
Add tire radius calculation utility in Rust

kylecarow:


rust/fastsim-core/src/utils.rs
```diff
@@ -294,6 +296,51 @@ pub fn interp2d(
...
 
+lazy_static! {
+    static ref TIRE_CODE_REGEX: Regex = Regex::new(
+        r"(?i)[P|LT|ST|T]?((?:[0-9]{2,3}\.)?[0-9]+)/((?:[0-9]{1,2}\.)?[0-9]+) ?[B|D|R]?[x|\-| ]?((?:[0-9]{1,2}\.)?[0-9]+)[A|B|C|D|E|F|G|H|J|L|M|N]?"
+    ).unwrap();
+}
+
+/// Calculate tire radius (in meters) from an [ISO metric tire code](https://en.wikipedia.org/wiki/Tire_code#ISO_metric_tire_codes)
+///
+/// **Example 1:**
```
calbaker: @kylecarow , does `**Example 1:**` do anything signifiant in markdown?  If not, I'd recommend changing it to something like
```
# Examples
## Example 1 
...```

kylecarow: It bolds text
\*\*example\*\* becomes **example**

I'm fine with either really
image

image

calbaker (approved):

PR #69, opened 2023-11-08
remove some redundancies

kylecarow:
Removed some redundant code and fields. Serde still loads old simdrive YAMLs that have the removed fields without issue, they are just ignored on loading.


kylecarow:
@calbaker this also removes the goofy code snippets that I shared with you earlier today

python/fastsim/simdrive.py
```diff
@@ -948,14 +940,14 @@ def set_power_calcs(self, i):
 
             if self.veh.fc_eff_type == H2FC:
                 self.min_mc_kw_2help_fc[i] = max(
-                    self.trans_kw_in_ach[i], -self.cur_max_mech_mc_kw_in[i])
+                    self.trans_kw_in_ach[i], -self.ess_lim_mc_regen_kw[i])
```
kylecarow: @calbaker Replacing this redundant variable makes the name kind of ambiguous, so I'd be cool with changing it to something more descriptive (for example, if we changed it back to `cur_max_mech_mc_kw_in` or something)

kylecarow:
@calbaker Do we need `RustCycle::get_len()` in addition to `RustCycle::len()`?

python/fastsim/simdrive.py
kylecarow: can just use old name

kylecarow:
> @calbaker Do we need `RustCycle::get_len()` in addition to `RustCycle::len()`?

can remove the `.len` getter

kylecarow:
This is ready for review (cyc len change is in [different PR](https://github.com/NREL/fastsim/pull/75))

calbaker (approved):

PR #71, opened 2023-11-22
Remove unnecessary clones

kylecarow:


kylecarow:
get rid of `RustCycle::new` and implement from dict (and from hashmap)

PR #72, opened 2023-11-22
Use ndarray stats

kylecarow:
This PR eliminates `ndarrmax` and `ndarrmin` and instead uses the methods implemented by [ndarray_stats::QuantileExt](https://docs.rs/ndarray-stats/latest/ndarray_stats/trait.QuantileExt.html)

For example:
```diff
- let mc_max_eff = ndarrmax(&self.mc_eff_array);
+ let mc_max_eff = self.mc_eff_array.max().unwrap();
```

### DIFF
--- a/rust/fastsim-core/Cargo.toml
+++ b/rust/fastsim-core/Cargo.toml
@@ -33,6 +33,7 @@ lazy_static = "1.4.0"
 regex = "1.7.1"
 rayon = "1.7.0"
 include_dir = "0.7.3"
+ndarray-stats = "0.5.1"
 
 [package.metadata]
 include = [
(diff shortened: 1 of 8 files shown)


kylecarow:
this might be useful for other features, see the docs here: https://docs.rs/ndarray-stats/latest/ndarray_stats/
but otherwise this can be reviewed & merged

kylecarow:
@calbaker this should be easy to review, thought it would be nice using this crate's `impl` on `ndarray::ArrayBase` to provide `min` and `max` functions rather than our own versions

kylecarow:
@calbaker @robinsteuteville related to our discussion today: it seems that Rust only compiles code in dependencies that is actually used!
https://www.reddit.com/r/rust/comments/7z5z6h/dead_code_elimination_in_rust/

PR #73, opened 2023-11-22
replace interp2d with n-dimensional interpolator

kylecarow:


kylecarow:
could probably parallelize finding surrounding indices in each dimension

calbaker (approved):

PR #77, opened 2023-12-06
Fix simdrivelabel discrepancy for EVs

kylecarow:
- [x] fix bug with simdrivelabel EVs
- [ ] enhance tests for simdrivelabel so that every vehicle type is tested


kylecarow:
@calbaker I cherry-picked the bug fix into `fastsim-2` by the way, so this PR should just be your test changes now.

PR #78, opened 2023-12-11
Simdrivelabel EV fixes

kylecarow:


jakeholden:
@kylecarow this looks good to me, is there any particular reason that only some of these how their own individual model file?

jakeholden (approved):

kylecarow:
> @kylecarow this looks good to me, is there any particular reason that only some of these how their own individual model file?

This is somewhat hard to explain over text, I'll explain in person when I'm in the office tomorrow.

PR #59, opened 2023-10-20
add init calls in `from_*`methods

kylecarow:

PR #83, opened 2023-12-18
Add init calls in SerdeAPI trait methods

kylecarow:
Replaces https://github.com/NREL/fastsim/pull/59

- `SerdeAPI` docstring enhancements
- all `SerdeAPI::from_...` methods call `init` now
- fixed bugs that popped up
  - only check mc eff bounds if `!no_elec_sys`, this turned out to not be strictly necessary for tests to pass, but no sense in checking it for `veh_pt_type == Conv`
  - `to_rust=True` resulted in `modern_max` defaulting to zero instead of setting the parameter, fixed by uncommenting this line: https://github.com/NREL/fastsim/blob/a1ea3359804ad90db09578f0e8f4d71f32cf6e8e/rust/fastsim-core/src/vehicle.rs#L752
  ~actually, do we want this to only be set if `self.modern_max` is zero?~ did this
- added `initialized: bool` field for `RustCycle` and `RustVehicle` to not rerun `SerdeAPI::init` if `self.initialized == true`


kylecarow:
@calbaker This is ready for review. Doesn't feature the `initialized` field I proposed, which can be a follow-on PR, or I could add that if you prefer.

kylecarow:
Actually, going to switch this to draft and add the `initialized` field, I tested it out and it should be pretty easy

kylecarow:
@calbaker I'm actually having second thoughts about the `initialized` field... I have rewritten this so that `init` only gets called in `from_reader`, and the methods that `SerdeAPI::from_str` calls (`from_yaml`, `from_json`). This covers all `from_...` functions in the codebase, and only calls `init` once. So, I don't think this field is actually very necessary.

So it seems like this `initialized` field check can be totally removed, with one very small caveat. I would also remove the `init` call that I added in `RustCycle::push`: https://github.com/NREL/fastsim/blob/2f4f4339b14bfa3d486575f7e9246ecaa44eff44/rust/fastsim-core/src/cycle.rs#L846-L862

Meaning if someone just uses `push` to add a single element it wouldn't run `init`. If we kept the `init` there, it would mean `RustCycle::from_reader` for the csv case would run `init` every time an element is added, see line 728 here: https://github.com/NREL/fastsim/blob/2f4f4339b14bfa3d486575f7e9246ecaa44eff44/rust/fastsim-core/src/cycle.rs#L722-L732

I think this small case is not enough to warrant `initialized: bool` at all. Can you think of any other reason to keep it?


Made a diagram showing the call structure of SerdeAPI
image

calbaker (approved):

PR #86, opened 2023-12-19
adding new from_github vehicle method

robinsteuteville:


rust/fastsim-core/src/cycle.rs
```diff
@@ -726,6 +726,19 @@ impl SerdeAPI for RustCycle {
             ),
         }
     }
+
+    /// takes an object from a url and saves it in the fastsim data directory in a rust_objects folder
+    /// WARNING: if there is a file already in the data subdirectory with the same name, it will be replaced by the new file
+    fn to_cache<S: AsRef<str>>(url: S) {
```
calbaker: @robinsteuteville add this:
```rust
/// # Arguments
/// - url: string or path to url containing model
```

and format other doc strings similarly.   Your description is good.  I'd probably change `rust_objects` to `cycles`  in the doc string.

rust/fastsim-core/src/cycle.rs
```diff
@@ -726,6 +726,19 @@ impl SerdeAPI for RustCycle {
             ),
         }
     }
+
+    /// takes an object from a url and saves it in the fastsim data directory in a rust_objects folder
```
calbaker: make sure there is a trailing double space, `  ` on this line so that the following line is actually on the next line when rendered.  You can check rendering by running `cargo doc --no-deps --open`.

rust/fastsim-core/src/cycle.rs
```diff
@@ -726,6 +726,19 @@ impl SerdeAPI for RustCycle {
...
+    /// WARNING: if there is a file already in the data subdirectory with the same name, it will be replaced by the new file
```
calbaker: This warning is good, and I prefer the behavior you have now and this warning being in the doc string as opposed to adding another function args.

rust/fastsim-core/src/lib.rs
```diff
@@ -29,6 +29,7 @@
 //! ```
 
 extern crate ndarray;
+extern crate tempfile;
```
calbaker: you probably don't need this line.  The `extern` syntax has been unneeded since a much earlier version of rust.  See what happens if you delete it.

rust/fastsim-core/src/traits.rs
```diff
@@ -154,6 +155,44 @@ pub trait SerdeAPI: Serialize + for<'a> Deserialize<'a> {
...
     }
+
+    /// instantiates an object from a url
+    fn from_url<S: AsRef<str>>(url: S) -> anyhow::Result<Self> {
+        let url = url.as_ref();
+        let temp_dir = tempdir()?;
+        let mut file_path = PathBuf::new();
+        // do these file types need to be specific to the object?
+        // TODO: either make funciton work for csv files, or remove from supported file list
+        if url.ends_with("yaml"){
```
calbaker: This totally works, but see if you can figure out how to represent this `if ... else` chain with a `match` representation similar to https://github.com/NREL/fastsim/blob/5d2c1d4b331c58a87c9bd390e1ff6005bc22e8ca/rust/fastsim-core/src/traits.rs#L22.

calbaker:
@robinsteuteville it might be good to use on or both of [these constants](https://github.com/NREL/fastsim/blob/5d2c1d4b331c58a87c9bd390e1ff6005bc22e8ca/rust/fastsim-core/src/traits.rs#L6).  You might also find that you need to move them out of the trait definition, which is ok to do if needed.

rust/fastsim-core/src/traits.rs
```diff
@@ -154,6 +155,44 @@ pub trait SerdeAPI: Serialize + for<'a> Deserialize<'a> {
...
+            let file_path = temp_dir.path().join("temporary_object.yaml");
+        } else if url.ends_with("csv"){
+            let file_path = temp_dir.path().join("temporary_object.csv");
+        } else if url.ends_with("json"){
+            let file_path = temp_dir.path().join("temporary_object.json");
+        } else {
+            bail!("Unsupported file type, must be a yaml, json, or csv file.");
```
calbaker: Make this bail message look more like what's here:
https://github.com/NREL/fastsim/blob/5d2c1d4b331c58a87c9bd390e1ff6005bc22e8ca/rust/fastsim-core/src/traits.rs#L73

rust/fastsim-core/src/vehicle_utils.rs
```diff
@@ -1771,6 +1771,33 @@ pub fn import_and_save_all_vehicles(
...
+pub fn fetch_github_list() -> anyhow::Result<Vec<String>> {
+    let response = ureq::get(VEHICLE_REPO_LIST_URL).call()?.into_reader();
+    let github_list: HashMap<i64, GitVehicleInfo> = serde_json::from_reader(response).with_context(||"Cannot parse github vehicle list.")?;
+    let mut vehicle_name_list: Vec<String> = Vec::new();
+    for (key, value) in github_list.iter() {
+        vehicle_name_list.push(value.name.to_owned())
+    }
+    Ok(vehicle_name_list)
+}
+
```
kylecarow, line 1868: cool!

rust/fastsim-core/src/vehicle_utils.rs
```diff
@@ -1771,6 +1773,99 @@ pub fn import_and_save_all_vehicles(
...
+    let github_list: Vec<GitObjectInfo> =
+        serde_json::from_reader(response).with_context(|| "Cannot parse github vehicle list.")?;
+    let mut vehicle_name_list: Vec<String> = Vec::new();
+    for object in github_list.iter() {
+        if object.url_type == "dir" {
+            let url = &object.url;
+            let vehicle_name_sublist = fetch_github_list(Some(url.to_owned()))?;
+            for name in vehicle_name_sublist.iter() {
+                vehicle_name_list.push(name.to_owned());
+            }
```
kylecarow, line 1847: @robinsteuteville Would this only look one level deep into directories?

robinsteuteville: @kylecarow I believe it should look as many levels deep into directories as needed, since it calls fetch_github_list recursively within itself, which should mean that each time it comes across a subdirectory within the current directory it is in, it should start the process over again for that subdirectory, which in turn will uncover any subsubdirectories, etc. Does make my brain hurt a bit, so if you see a problem within the recursive logic let me know, I definitely could be thinking about it wrong!

rust/fastsim-core/Cargo.toml
```diff
@@ -39,6 +39,9 @@ include_dir = "0.7.3"
 itertools = "0.12.0"
 ndarray-stats = "0.5.1"
 tempfile = "3.8.1"
+url = "2.5.0"
```
calbaker, line 42: @kylecarow, could you feature gate this stuff?

rust/fastsim-core/src/cycle.rs
```diff
@@ -631,6 +631,8 @@ pub struct RustCycle {
 impl SerdeAPI for RustCycle {
     const ACCEPTED_BYTE_FORMATS: &'static [&'static str] = &["yaml", "json", "bin", "csv"];
     const ACCEPTED_STR_FORMATS: &'static [&'static str] = &["yaml", "json", "csv"];
+    // is this enough, or do I have to copy paste in the whole to_cache mathod?
```
calbaker:
```suggestion
```

kylecarow (approved):

PR #89, opened 2024-01-03
updating vehicle_import_demo

robinsteuteville:
Updating vehicle_import_demo.py to utilize new demo testing.


python/fastsim/__init__.py
```diff
@@ -9,7 +9,7 @@
 from . import fastsimrust
 from . import fastsimrust as fsr
 from . import parameters as params
-from . import utilities as utils
+from .utils import utilities as utils
```
kylecarow: @robinsteuteville 
Try `from . import utils` here and in utilities/__init__.py add `from .utilities import *` and check out the effect :)

kylecarow: ... and in utilities/__init__.py add `from .utilities import *` and check out the effect :)

kylecarow: It's subtle, but this doesn't change anything about the package's namespace, and allows other utilities in the directory to be exposed as well (like if we wanted to expose the other file's functions under `fastsim.utils`, we could just edit utilities/__init__.py to import them as well)

robinsteuteville: updated!

kylecarow:
@calbaker This looks good to me!

kylecarow:
@calbaker I think we reviewed this in the FASTSim dev meeting, is this good to merge?

calbaker (approved):

PR #90, opened 2024-01-03
from_resource now calls from_reader

kylecarow:
Also `from_csv_str` now takes `name: String`, which is totally fine as it allocates anyway, so the allocation happens explicitly within the caller rather than (possibly) multiple times.


kylecarow:
Merging because I have tested this extensively

PR #93, opened 2024-01-18
add volvo c40 to fastsim-2

kylecarow:
See https://github.com/NREL/fastsim/pull/91


kylecarow:
@jakeholden Volvo C40 is now in FASTSim 2

PR #94, opened 2024-01-22
refactor assignment syntax in simdrive_impl

kylecarow:
Noticed the syntax on assignments to simdrive arrays could be cleaner, so I tidied those up.

Before:
```rust
if condition_1 {
    LHS = value_1;
} else if condition_2 {
    LHS = value_2;
} else {
    LHS = value_3;
}
```
After:
```rust
LHS = if condition_1 {
    value_1
} else if condition_2 {
    value_2
} else {
    value_3
};
```

Also prevented recalculating `dt_s()` many times in `set_post_scalars`


calbaker:
> Also prevented recalculating `dt_s()` many times in `set_post_scalars`

Does this mean it's faster?

calbaker (approved):

kylecarow:
> Does this mean it's faster?

@calbaker I would assume so, but I didn't benchmark it. It's also in a function that's only run once per `sim_drive()` call

PR #95, opened 2024-01-24
implement to_writer method and clean up RustCycle csv serde

kylecarow:
- adds a `to_writer` method
  - RustCycle just needs to override `to_writer` instead of `to_file`
  - Cleanup/reorg of SerdeAPI-alike functions in RustCycle (`write_csv` etc)
  - New `RustCycle::to_csv` function to be consistent with `to_yaml` etc, same functionality as `RustCycle::to_str("csv")` (which now calls `RustCycle::to_csv`)
- All SerdeAPI errors are now exposed as `IOError`/`OSError` in Python rather than `RuntimeError`

I noticed the Python-exposed `RustCycle::from_csv` loads a cycle from a given filepath, rather than a string like `from_yaml`, `from_json`, etc. I think this is something that I would definitely want to change in FASTSim 3. @calbaker what are your thoughts about making all these string deserialization methods have `_str` at the end (starting in FASTSim 3, to avoid API changes)?

i.e. `from_yaml` would become `from_yaml_str` to make it clear that it takes a YAML-formatted string, rather than a filepath

### DIFF
--- a/rust/fastsim-core/fastsim-proc-macros/src/add_pyo3_api/mod.rs
+++ b/rust/fastsim-core/fastsim-proc-macros/src/add_pyo3_api/mod.rs
@@ -215,8 +215,8 @@ pub fn add_pyo3_api(attr: TokenStream, item: TokenStream) -> TokenStream {
         ///
         #[staticmethod]
         #[pyo3(name = "from_resource")]
-        pub fn from_resource_py(filepath: &PyAny) -> anyhow::Result<Self> {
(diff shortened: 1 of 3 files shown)


calbaker:
@kylecarow , is this ready to merge?

kylecarow:
@calbaker Left a comment for you to ponder, but this PR is ready to merge, yea

calbaker (approved):

PR #102, opened 2024-01-29
Cargo features

calbaker:
creates optional `full` feature (enabled by default) to allow for optional reduced-dependency version of fastsim


calbaker:
Converted to draft pending review with Google.

rust/fastsim-py/Cargo.toml
```diff
@@ -12,7 +12,9 @@ repository = "https://github.com/NREL/fastsim"
 
 [dependencies]
 pyo3-log = { workspace = true, optional = true }
-fastsim-core = { path = "../fastsim-core", features = ["pyo3"], version = "~0" }
+fastsim-core = { path = "../fastsim-core", features = [
+    "pyo3",
+], version = "~0", default-features = false }
```
kylecarow, line 17: @calbaker see this change

calbaker: But this means that `fastsim-core` is always compiled with default features off rather than getting propagated from the maturin command.   There should be a way to do that.  @kylecarow

kylecarow: Since fastsim-py has a default "full" feature that activates "fastsim-core/full", I don't think there is a functional difference.

calbaker: @kylecarow, that sounds right.  I'll try to convince myself of it tomorrow.

kylecarow:
@calbaker could we look at merging this? There are changes to vehicle import organization that I want to work on in fastsim-2 without needing to wait for abc to drag coeff organization (the last remaining thing feature-gated only by the `default` feature)

PR #104, opened 2024-01-31
fixing python documentation formatting

robinsteuteville:


robinsteuteville:
issue: https://github.nrel.gov/MBAP/fastsim/issues/296

robinsteuteville:
@calbaker I was able to fix the `copy_sim_drive` documentation formatting, although I had to add a line of description in order to do so -- the line can be changed or updated as needed. I figured out two things. One, from what I can tell it seems like in order for an argument to be recognized as such, it has to have a description on the same line as the argument (so, argument: xxxxxxx, all on one line). Two, any additional lines of description for an argument (like the bullet points in `copy_sim_drive`) must align with the beginning of the first line of description. Here's the updated docstring for `copy_sim_drive`:
Screenshot 2024-01-31 at 10 15 09 AM
And here's the resulting format:
Screenshot 2024-01-31 at 10 15 46 AM

PR #105, opened 2024-02-02
updating github tests

robinsteuteville:
issue: https://github.nrel.gov/MBAP/fastsim/issues/294


robinsteuteville:
issue: https://github.nrel.gov/MBAP/fastsim/issues/294

robinsteuteville:
Note: `contains(github.event.pull_request.changed_files, 'rust/'))` seems to only check the most recent commit -- not all changes on the branch. This could be more problematic than helpful. Not sure whether it is possible to make this check all changes on the branch rather than just the most recent commit.

PR #110, opened 2024-02-19
fixing test_output folder and related issues

robinsteuteville:


robinsteuteville:
@calbaker @kylecarow I think this will need more troubleshooting before it is ready to merge. I thought it was ready, but then I noticed that the github tests took way longer than usual to run, so I'm gonna try to figure why.

calbaker (approved):

PR #113, opened 2024-02-28
attempt to fix -e in actions

kylecarow:


calbaker:
@kylecarow, what's the deal with this PR?

.github/workflows/tests.yaml
```diff
@@ -41,5 +41,5 @@ jobs:
       # https://github.nrel.gov/MBAP/fastsim/issues/310
       - name: Python unit tests 
         run: |
-          pip install -e ".[dev]" && pytest -v python/fastsim/tests/
```
calbaker, line 44: @kylecarow looks like maybe we should undo these changes in a new commit, change the PR description from `attempt to fix -e in actions` to `failed attempt to fix -e in actions with other minor patch changes`

kylecarow:
> @kylecarow, what's the deal with this PR?

This explains it https://github.nrel.gov/MBAP/fastsim/issues/310

Installing the package will go faster if it is non-editable

PR #118, opened 2024-04-23
updating fastsim-3 readme

robinsteuteville:


robinsteuteville:
@calbaker there is no acknowledgements section of the `fastsim-2` readme, only a contributors section.
For the `FASTSim-3` installation section, what needs to be changed for the installation process for Rust and Python?

README.md
```diff
@@ -43,90 +64,9 @@ Developers might want to install the code in place so that FASTSim files can be
...
-# New Features and Capabilities
-In `fastsim-3`, we introduce:
-- modular, hierchical component structure 
-- component-specific vehicle models -- i.e. the conventional vehicle contains only data for the fuel
-  converter and other relevant components but does contain any dummy battery or motor parameters as
-  is the case in `fastsim-2`
-- faster run times due to modularity and extensive streamlining
-- ability to control granularity of time-resolved data -- e.g. save at every time step, save at every _n_th time step, or never save at all (saving only cumulative trip-level results)
-- ability to simulate standalone component models
+To see and run examples, navigate to `./python/fastsim/demos` and run the various *demo.py files
```
calbaker: @robinsteuteville , could you update this section to show how to copy the demo files locally using the function we made?  This will also need to happen in `fastsim-2` I think.

calbaker:
@robinsteuteville, put your name in https://github.com/NREL/fastsim/blob/8ebd695d897e58cdee37661444a961bb1b402e99/README.md#contributors and the equivalent location in fastsim-2

README.md
robinsteuteville: @calbaker 
I will definitely add directions in the `fastsim-2` readme.
The function in `fastsim-2` that copied the demos was called `copy_demo_files`. I can't find a similar function in `fastsim-3`. Is it under a different name, or does it not exist yet? If it doesn't exist yet, should we wait until it does exist before writing instructions into the `fastsim-3` readme?

robinsteuteville:
@calbaker in line 51 of the `README.md` file, where it gives instructions for how to download `fastsim-3`, is `pip install fastsim` the correct way to install `fastim-3`, or does the user need to specify somehow that they want `fastsim-3` downloaded and not `fastsim-2`?

calbaker (approved):

PR #119, opened 2024-04-26
Feature/log level input checking

calbaker:


python/fastsim/demos/demo.py
```diff
@@ -482,92 +482,94 @@ def get_sim_drive_vec(
...
-
-    cyc = {}
-    cyc['cycGrade'] = np.zeros(len(pnts))
-    cyc['mps'] = np.array(
-        pnts['speed_mph'] / fsim.params.MPH_PER_MPS)  # MPH to MPS conversion
-    cyc['time_s'] = np.array(
-        np.cumsum(
-            (pnts['time_local'] -
-             pnts['time_local'].shift()).fillna(pd.Timedelta(seconds=0)).astype('timedelta64[s]')
+with fsim.utils.suppress_logging():
```
kylecarow, line 485: Can we change instances of `fsim.utils.suppress_logging():` to a single `fsim.utils.disable_logging()` at the top of the file?

python/fastsim/utils/utilities.py
```diff
@@ -117,6 +117,28 @@ def set_log_level(level: str | int) -> int:
...
+        # no logging of anything ever!
+        logging.CRITICAL + 1,
+    ]
+
+    err_str = f"Invalid arg: '{level}'.  See doc string:\n{set_log_level.__doc__}"
+
+    assert level in allowed_args, err_str
+
     if isinstance(level, str):
         level = logging._nameToLevel[level]
```
kylecarow, line 143: @calbaker optionally:
```suggestion
        level = logging._nameToLevel[level.upper()]
```
This would allow lowercase log level strings `"warning"`, `"info"`, etc., but we would need to add those to the allowable inputs.

kylecarow (changes requested):

python/fastsim/demos/demo.py
calbaker: @kylecarow, I prefer the latter because I'd like it to be really obvious that we're suppressing logging so that we remember to fix the cause of the logging at some point in the future.  Does this seem ok?

python/fastsim/utils/utilities.py
calbaker: @kylecarow , yea that sounds good.  Could you make that change?

python/fastsim/demos/demo.py
kylecarow: I'm fine with either way really. Do you mean keeping it as is?

calbaker:
moved to https://github.com/NREL/fastsim/pull/126

PR #129, opened 2024-05-21
F3/hev

calbaker:
- Adds PHEV model that is tested in rust
- TODO: write a demo script in python for the hev and compare to fastsim-2 hev results


michael-okeefe:
@calbaker I'm seeing the CI checks / tests aren't passing; I just re-requested they run again in case something was stale but I'm getting the same result.

michael-okeefe:
@calbaker Should the python unit test section of the workflow be removed or commented out, then?
See bottom of [the CI tests workflow file](https://github.com/NREL/fastsim/blob/f3/hev/.github/workflows/tests.yaml)

PR #131, opened 2024-05-27
updates the fastsim-2 demo files to run Rust instead of Python

robinsteuteville:
This commit updates the `fastsim-2` demo files to run Rust instead of Python where Python was formerly used for `sim_drive` functions, and only use Rust where Rust and Python were both used (except in the one case where Rust and Python speeds are being intentionally and directly compared (code lines 50-116 of [demo.py](https://github.com/NREL/fastsim/blob/have-demos-run-rust/python/fastsim/demos/demo.py)).

### DIFF
--- a/python/fastsim/demos/timing_demo.py
+++ b/python/fastsim/demos/timing_demo.py
@@ -56,7 +56,7 @@ print(f'Time to load cycle with `fastsimrust.RustCycle.from_file`: {t1 - t0:.2e}
 
 veh_csv_path = fsim.package_root() / "resources/vehdb/2012_Ford_Fusion.csv"
 t0 = time.perf_counter()
-veh_python = fsim.vehicle.Vehicle.from_file(str(veh_csv_path), to_rust=True)
+veh_python = fsim.vehicle.Vehicle.from_file(str(veh_csv_path), to_rust=True).to_rust()
 print(
     'Time to load vehicle with\n' + 
     f'`vehicle.Vehicle.from_file(veh_csv_path)`: {time.perf_counter() - t0:.2e} s'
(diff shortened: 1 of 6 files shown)


calbaker (approved):

python/fastsim/demos/timing_demo.py
```diff
@@ -56,7 +56,7 @@
 
 veh_csv_path = fsim.package_root() / "resources/vehdb/2012_Ford_Fusion.csv"
 t0 = time.perf_counter()
-veh_python = fsim.vehicle.Vehicle.from_file(str(veh_csv_path), to_rust=True)
+veh_python = fsim.vehicle.Vehicle.from_file(str(veh_csv_path), to_rust=True).to_rust()
```
kylecarow, line 59: @calbaker @robinsteuteville I would think this line should be:
`veh_python = fsim.vehicle.Vehicle.from_file(str(veh_csv_path))`

Am I misunderstanding something?

calbaker: @kylecarow, `fsim.vehicle.Vehicle.from_file` uses the python-defined code, and the `to_rust` method converts it to a rust object.  The goal was to have the demo files run rust rather than running the slower python so that `build_and_test.sh` runs faster.  @robinsteuteville , right?

kylecarow: Ah, I see, I think I was misunderstanding this part of the demo - both lines instantiate `RustVehicle`s. I think we ought to update the benchmark results starting on line 51 to reflect the new code.

calbaker: @kylecarow, could you see what needs to be done to make this PR mergeable?

kylecarow: @calbaker I reviewed all changes and seems that it's ready to merge as-is

kylecarow (approved):

calbaker:
@kylecarow does this get us through unit tests any faster?  If not, got any ideas?

kylecarow:
@calbaker Is there a good way to test this, given that it's been non-squash merged? I would've directly compared branches before the merge, but after the merge there are mixed commits up to 4 months ago

If not we can revert the merge to test this

kylecarow:
My demo tests went from ~29 seconds to ~20 seconds, nicely done @robinsteuteville!!

PR #132, opened 2024-05-30
Add copy_demo_files()

robinsteuteville:
I added copy_demo_files() into `fastsim-3`, as well as adding a `test_utils.py` file with a `test_copy_demo_files()` function that runs during `sh build_and_test.sh`. I tested the function manually as well, and it correctly copies the current `fastsim-3` demos into either a `demos` folder or a specified path.

### DIFF
--- a/python/fastsim/__init__.py
+++ b/python/fastsim/__init__.py
@@ -9,6 +9,8 @@ def package_root() -> Path:
     """Returns the package root directory."""
     return Path(__file__).parent
 
+from pkg_resources import get_distribution
+__version__ = get_distribution("fastsim").version
 
 def set_param_from_path(
     model: Any,
--- a/python/fastsim/tests/test_utils.py
+++ b/python/fastsim/tests/test_utils.py
@@ -1,0 +1,21 @@
+import tempfile
+import unittest
+from pathlib import Path
+import fastsim as fsim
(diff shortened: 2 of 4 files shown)


robinsteuteville:
@calbaker found a small thing I need to update so it's not quite ready to merge yet.

robinsteuteville:
@calbaker this PR is ready for review -- I learned the hard way that if you switch between versions FASTSim within the same vscode session, the `fsim.__version__` does not seem to automatically change -- I had to close out my session and start a new one. So, if you use one version of FASTSim then switch to a different version in the same session, and then download the demos for the new version of FASTSim, the string appended to the beginning of the downloaded demos may list the wrong version (although the demos themselves will be the correct version). So, for instance I switched from `fastsim-2` to `fastsim-3` in the same session, and the `fastsim-3` demos I downloaded appended the string `# %% Copied from FASTSim version '{v}'. Guaranteed compatibility with this version only.` where `{v}` was a `fastsim-2` version rather than a `fastsim-3`. Once I restarted my session, the version was listed correctly as `3.0.0`. Besides that, the `copy_demo_files()` function works as intended to download the proper demos to a `demos` folder or a specified folder, and works regardless of whether the argument is a string or a Path.

calbaker (approved):

PR #136, opened 2024-06-11
Generate variable paths

robinsteuteville:


robinsteuteville:
@cbaker2 I created a unit test testing the function output against what I think is the complete output for a fastsim Vehicle. Does the list I'm comparing the output to look good to you, or is it missing anything (the test passes, so if there is something missing, that means the function is not working properly yet)?

python/fastsim/demos/demo_variable_paths.py
```diff
@@ -0,0 +1,27 @@
+"""
+Script demonstrating how to use variable_path_list() and history_path_list()
+demos to find the paths to variables within fastsim classes.
+"""
+
+import fastsim as fsim
+
+# load 2012 Ford Fusion from file
+veh = fsim.Vehicle.from_file(
+    str(fsim.package_root() / "../../tests/assets/2016_TOYOTA_Prius_Two.yaml")
```
calbaker, line 10: @robinsteuteville , switched to the prius to see if it'll work on HEVs, and there is an error with `None`, perhaps the same as what you encountered in ALTRIOS.

calbaker:
superseded by #147

PR #141, opened 2024-06-27
Add docstrings to RustSimDriveParams

michael-okeefe:
Fixes: https://github.nrel.gov/MBAP/fastsim/issues/348

The docstrings were added by referencing other parts of the code. Some were slightly adjusted to enhance clarity where possible.

NOTE: this also incorporates a change that seems to have fixed the unit tests not building in Github actions:

The root cause *appears* to be a breaking change in numpy 1.24 and later that causes an error in another dependency (pymoo?). I limited the max version of numpy to be <1.24 to fix. A more comprehensive fix might involve updating all the deps.


michael-okeefe:
NOTE: I noticed the unit tests don't seem to be passing via Github Actions. I tried an experiment based on this thread to increase the swap space thinking it might have been a memory issue but that didn't work:

https://github.com/actions/runner/issues/2468

Not sure at this point what is wrong with the automated tests although I wonder if it could be the "vehicle import" tooling since that requires network access? Just a wild guess at this point...

michael-okeefe:
OK, not sure if this definitively fixed the unit tests but it appears to be a breaking change in numpy 1.24 and later that causes this. This shows up in one of the dependencies. I limited the max version of numpy to be <1.24 to fix but there's probably a better, more comprehensive fix involving updating all the deps.

rust/fastsim-core/src/simdrive.rs
```diff
@@ -19,37 +19,75 @@ pub mod simdrive_iter;
 )]
 /// Struct containing time trace data
 pub struct RustSimDriveParams {
+    /// if true, accuracy will be favored over performance for grade per step estimates
+    /// Specifically, for performance, grade for a step will be assumed to be the grade
+    /// looked up at step start distance. For accuracy, the actual elevations will be
+    /// used. This distinciton only makes a difference for CAV maneuvers.
     pub favor_grade_accuracy: bool,
-    pub missed_trace_correction: bool, // if true, missed trace correction is active, default = false
```
calbaker, line 23: @jakeholden , check out these doc strings

calbaker:
Also addresses [#9](https://github.nrel.gov/MBAP/covar-eco-routing/issues/9)

jakeholden (approved):
These new docs strings look excellent!!

calbaker:
@jakeholden , @michael-okeefe should get most of the credit for this, FYI

PR #142, opened 2024-07-10
fix interpolation unwrap

calbaker:
fixes error described in commit message in dc22c077

Since #143 merges all this into `fastsim-3`, this PR exists solely to make it really easy to get feedback from Kyle on the changes introduced here.

### DIFF
--- a/fastsim-core/src/utils.rs
+++ b/fastsim-core/src/utils.rs
@@ -144,7 +144,7 @@ pub fn interp1d(
         if x >= &x_data[size - 2] {
             i = size - 2;
         } else {
-            while x > &x_data[i + 1] {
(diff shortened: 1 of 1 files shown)


fastsim-core/src/utils.rs
```diff
@@ -144,7 +144,7 @@ pub fn interp1d(
         if x >= &x_data[size - 2] {
             i = size - 2;
         } else {
-            while x > &x_data[i + 1] {
+            while i < x_data.len() - 2 && x > &x_data[i + 1] {
```
calbaker, line 174: @kylecarow , could you verify that this is reasonable? See the linked error.

PR #146, opened 2024-07-24
F3/interp with benchmarking

calbaker:
- feature gated logging to fix performance problems.  We're targeting 10x faster than fastsim-2, but with inactive, compiled logging, we get about 0.7x of fastsim-2 speed.  After feature gating, we're about 8x faster when not saving and 4x faster when saving
 - wrote but did not use macro that should be able to log function duration
 - this PR might include everything in #127, but is not meant to usurp that PR, which should proceed when @robinsteuteville  finishes applying the `Interpolator` everywhere.

- [ ] finalize when #127 is complete


calbaker:
superseded by #150

PR #147, opened 2024-07-29
F3/generate variable paths

calbaker:
replaces #136 with more descriptive branch name

solves [#345](https://github.nrel.gov/MBAP/fastsim/issues/345)

# TODO
- [x] change tack to use a dict rather than `dir()` for traversing the hierarchy
- [x] remove `test_variable_paths`
- [x] add an assert to check that the `variable_path_list` output in demo_variable_paths.py matches the expected result stored in test assets
- [x] add an assert to check the shape of the `sd.to_dataframe()` or maybe even load the dataframe from file to compare
- [x] `with_capacity` might eek out some better speedup for the `Some(<some int>)` case and the `None` case because it'd be pre-allocating a reasonable size or not pre-allocating at all, respectively -- no obvious speedup
~~- [ ] see if speed ratio can be bumped back up to 8 and 3~~


calbaker:
@robinsteuteville , please check if this does what you'd expect

python/fastsim/tests/test_variable_paths.py
```diff
@@ -0,0 +1,21 @@
+import unittest
+import fastsim as fsim
+
+class TestParamPath(unittest.TestCase):
```
calbaker: @robinsteuteville , please check if this test is doing what you'd expect.

python/fastsim/__init__.py
```diff
@@ -73,4 +96,139 @@ def _get_list(path_elem, container):
...
+
+@classmethod
+def from_pydict(cls, pydict: Dict) -> Self:
+    """
+    Instantiates Self from pure python dictionary 
+    """
+    import json
+    return cls.from_json(json.dumps(pydict))
+
+def to_dataframe(self, pandas:bool=False) -> [pd.DataFrame, pl.DataFrame, pl.LazyFrame]:
```
calbaker, line 205: @robinsteuteville , it might be good to make sure this is covered by a test.

PR #148, opened 2024-08-08
Add List Resources feature to FASTSim-2 (SerdeAPI tweaks, list_resources)

michael-okeefe:
This is the fastsim-2 implementation to address this issue:
https://github.nrel.gov/MBAP/fastsim/issues/354

The resources::list_resources(subdir) will list all the resources of a given subdirectory. If the subdirectory doesn't exist, an empty vector of string is returned, else the files in that subdirectory.

The SerdeAPI trait connects the RESOURCE_PREFIX with the call to resources::list_resources.

The list_resources_py methods of RustCycle and RustVehicle allow exposure from Python.

### DIFF
--- a/python/fastsim/tests/test_resources.py
+++ b/python/fastsim/tests/test_resources.py
@@ -1,0 +1,15 @@
+"""Test getting resource lists via Rust API"""
+import unittest
+
+import fastsim as fsim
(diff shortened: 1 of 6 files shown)


calbaker:
@michael-okeefe , this looks great!

PR #151, opened 2024-08-21
made diff generic and used it for dist

kylecarow:
Not sure that I implemented this in the best way, so check that out. It does a copy (via deref) on both values, there may be a better way.

### DIFF
--- a/fastsim-core/src/drive_cycle.rs
+++ b/fastsim-core/src/drive_cycle.rs
@@ -61,16 +61,17 @@ impl Init for Cycle {
         // somewhere to fix this ensure!(self.pwr_max_chrg.len() == self.len());
 
         // calculate distance from RHS integral of speed and time
-        self.dist = self
-            .time
-            .diff()
-            .iter()
-            .zip(&self.speed)
-            .scan(0. * uc::M, |dist, (time, speed)| {
-                *dist += *time * *speed;
-                Some(*dist)
-            })
-            .collect();
+        self.dist = {
+            let mut dt = vec![uc::S * 0.];
+            dt.extend(self.time.diff());
+            dt.iter()
+                .zip(&self.speed)
+                .scan(0. * uc::M, |dist, (dt, speed)| {
+                    *dist += *dt * *speed;
+                    Some(*dist)
+                })
+                .collect()
+        };
 
         // calculate elevation from RHS integral of grade and distance
         self.init_elev = self.init_elev.or(Some(get_elev_def()));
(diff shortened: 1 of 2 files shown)


calbaker:
@kylecarow , nicely done!  

@robinsteuteville, check out the cool trait stuff Kyle did in this PR.

nreinicke:
@kylecarow - Sorry I'm late looking at this but I'm wondering if you could implement the Diff trait by operating on cloned references instead of copying the underlying type?

```rust
pub trait Diff<T> {
    fn diff(&self) -> Vec<T>;
}

impl<T: Sub<Output = T> + Clone> Diff<T> for Vec<T> {
    fn diff(&self) -> Vec<T> {
        self.windows(2)
            .map(|vs| {
                let x = &vs[0];
                let y = &vs[1];
                y.clone() - x.clone()
            })
            .collect()
    }
}
```

Cloning the reference vs copying a float are probably similar performance wise but it might make a big difference if `T` is a bigger heap allocated type.

calbaker:
@nreinicke , when we have
```rust
    fn diff(&self) -> Vec<T> {
        let mut v_diff: Vec<T> = vec![Default::default()];
        v_diff.extend::<Vec<T>>(
            self.windows(2)
                .map(|vs| {
                    let x = &vs[0];
                    let y = &vs[1];
                    y.clone() - x.clone()
                })
                .collect(),
        );
        v_diff
    }

```

we get 
```
warning: using `clone` on type `T` which implements the `Copy` trait
   --> fastsim-core/src/traits.rs:393:21
    |
393 |                     y.clone() - x.clone()
    |                     ^^^^^^^^^ help: try dereferencing it: `*y`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#clone_on_copy
    = note: `#[warn(clippy::clone_on_copy)]` on by default

warning: using `clone` on type `T` which implements the `Copy` trait
   --> fastsim-core/src/traits.rs:393:33
    |
393 |                     y.clone() - x.clone()
    |                                 ^^^^^^^^^ help: try dereferencing it: `*x`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#clone_on_copy
```

nreinicke:
I guess this makes a good point in that if you're implementing this trait on a datatype that implements `Copy`, like `f64`, you're probably better off just doing it the original way (de-referencing) but if you're implementing this on a datatype that is heap allocated and expensive to clone, cloning by reference would be better.

PR #152, opened 2024-08-23
F3/serde with units

calbaker:
appends units to variable names when serializing and looks for (optional) units when deserializing


kylecarow:
Sweet! Are we also able to provide e.g. `pwr_aux_kilowatts` in place of `pwr_aux_watts`?

PR #155, opened 2024-09-13
Return numpy arrays from Rust

kylecarow:
@calbaker check this out!!!

I think setting using numpy arrays already worked, so this is something we can do now:
image


kylecarow:
It seems some pymethods that return `Vec`s (e.g. `RustCycle::get_dist_m`) should be changed to return numpy arrays as well, so they don't become lists on the Python side. @calbaker thoughts?

calbaker:
@kylecarow , is this ready to go?

kylecarow:
@calbaker The major thing that this needs is a warning or error on setting at an index: `my_numpy_array[i] = ...`
This PR doesn't do that yet

calbaker:
> @calbaker The major thing that this needs is a warning or error on setting at an index: `my_numpy_array[i] = ...` This PR doesn't do that yet

@kylecarow , could you maybe make a python class that inherits `np.array` and adds this behavior?

calbaker:
@kylecarow, ping!

kylecarow:
@calbaker This one is proving tricky, maybe we could look at it together sometime

PR #156, opened 2024-09-20
added a single vehicle to the `from_resources`

calbaker:
updated `list_resources` to live in the trait
removed `resources` module
`list_resources` ignores files with invalid extensions

@michael-okeefe  or @kylecarow , could either of you do a release of `fastsim-core` and Python fastsim when this passes?


rust/fastsim-core/src/traits.rs
```diff
@@ -1,7 +1,11 @@
-use crate::{imports::*, resources};
+use crate::imports::*;
+use include_dir::{include_dir, Dir};
 use std::collections::HashMap;
+use std::path::PathBuf;
 use ureq;
 
+pub const RESOURCES_DIR: Dir = include_dir!("$CARGO_MANIFEST_DIR/resources");
```
kylecarow, line 8:
```suggestion
#[cfg(feature = "resources")]
pub const RESOURCES_DIR: Dir = include_dir!("$CARGO_MANIFEST_DIR/resources");
```

rust/fastsim-core/src/traits.rs
```diff
@@ -1,7 +1,11 @@
-use crate::{imports::*, resources};
+use crate::imports::*;
+use include_dir::{include_dir, Dir};
```
kylecarow, line 2:
```suggestion
#[cfg(feature = "resources")]
use include_dir::{include_dir, Dir};
```

rust/fastsim-core/src/traits.rs
```diff
@@ -434,3 +452,23 @@ impl IterMaxMin<f64> for Array1<f64> {
         })
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use crate::imports::SerdeAPI;
+
+    #[test]
```
kylecarow, line 461:
```suggestion
    #[test]
    #[cfg(feature = "resources")]
```

PR #158, opened 2024-09-26
Use higher-quality logo file

kylecarow:
see README file


kylecarow:
@calbaker Also, we should probably skip the test workflow for non-code files

calbaker:
brilliant!  @kylecarow

calbaker (approved):

PR #166, opened 2024-11-14
Fix issues encountered during pytest

michael-okeefe:
Fixes some issues encountered during running `pytest -v` on Windows:

* updates how Python is called for demo tests so that it works within a virtual environment
* Handles how Polars DataFrame is printed by first transforming to a Pandas DataFrame

Additionally, this version adds a CI workflow file to build and test the application.


.github/workflows/build_and_test.yaml
```diff
@@ -4,34 +4,108 @@ on: [push]
...
       fail-fast: true
       matrix:
         os:
-          - ubuntu
-          - macos
-          - windows
+          - base: ubuntu
+            version: latest
+          - base: macos
+            version: '13'
```
calbaker, line 18: @michael-okeefe , could you drop some comments in here explaining why `latest` does not work for `macos` and why `13` and `14` are used here?  I want to make sure this is not hard to decipher in the future.

calbaker: @michael-okeefe , please also explain in the comments what event should trigger adding `15`, `16`, ...

calbaker:
> Handles how Polars DataFrame is printed by first transforming to a Pandas DataFrame
@michael-okeefe , where did this change happen?  I'm struggling to find it.

michael-okeefe:
> > Handles how Polars DataFrame is printed by first transforming to a Pandas DataFrame
> > @michael-okeefe , where did this change happen?  I'm struggling to find it.

@calbaker Apologies, I must have gotten my PRs mixed up as I don't see the change here, either. This was the area where the issue occurred during tests:

https://github.com/NREL/fastsim/blob/f3/fix-pytest-issues/python/fastsim/demos/demo_variable_paths.py#L49

Not sure if this is the exact line or not, though.

.github/workflows/build_and_test.yaml
michael-okeefe: Sure, I can do that. The issue is related to what processor each github CI machine is running for mac. MacOS 13 is Intel and MacOS 14 is Apple Silicon. Ideally, the Apple transition to their new chip occurs fast and we can drop having to call out specific machines and then just use latest, etc. 15, 16, etc. can be brought in at any time we desire and they should look like the calls for 14. As soon as 13 drops from the board, we can go back to just saying 'latest'.

michael-okeefe: @calbaker Note added to clarify why we are calling out macos-13 and 14.

.github/workflows/build_and_test.yaml
```diff
@@ -7,6 +7,9 @@ jobs:
     name: build py3.${{ matrix.python-version }} on ${{ matrix.platform || matrix.os.base }}-${{ matrix.os.version }}
     strategy:
       fail-fast: true
+      # NOTE: In the matrix below, we call out macos 13 and 14 explicitly. This is to enable running on both Intel and Apple Silicon for Mac.
+      # As of this writing, MacOS 13 is Intel and MacOS 14 is Apple Silicon. Ideally, the Apple transition to their new chip occurs fast and
+      # we can drop macos-13 checks and then just use 'latest'.
```
calbaker, line 12: @michael-okeefe , could we use `13` and `latest` so that we don't have to remember to increment the `14`?

michael-okeefe: @calbaker Great idea. I made the change and the tests are running.

calbaker (approved):

PR #168, opened 2024-11-15
adding set_default_xd_interp() methods for ReversibleEnergyStorage and adding Altrios battery maps

robinsteuteville:


fastsim-core/src/vehicle/powertrain/prepare_eff_values.py
```diff
@@ -0,0 +1,439 @@
+#%%
```
calbaker: What's `#%%` doing in here?  I think it'd be better to delete that.  

@robinsteuteville Could you put a doc string at the top of this file?  The very first thing in the file should be
```python
"""
this is is a single-line or multi-line doc string explaining why this file exists, 
what it does, and how to use it
"""

File: fastsim-core/src/vehicle/powertrain/efficiency_values.rs [hunk 2]
```diff
@@ -0,0 +1,225 @@
...
+    0.953873809588552,
+    0.927243266086586,
+    0.864025853742296,
+];
+
+// efficiency array as a function of power and SOC at constant 23 degrees C
+// corresponds to eta_interp_values[0] in Altrios, transposed so that the
+// outermost layer is now power, and the innermost layer SOC (in altrios,
+// the outermost layer is SOC and innermost is power)
+pub const F_XY_EFF: &[&[f64]] = &[
```
calbaker: @robinsteuteville 

For this large array, it'd be better to store it outside of rust with something like [include_str!](https://github.com/NREL/fastsim/blob/2f7e68971d088a4f56f6cca69dc7ca6646e317dd/fastsim-core/src/vehicle/vehicle_model.rs#L546) or [include_dir!](https://github.com/NREL/fastsim/blob/2f7e68971d088a4f56f6cca69dc7ca6646e317dd/fastsim-core/src/resources.rs#L2)

fastsim-core/src/vehicle/powertrain/prepare_eff_values.py
calbaker: @robinsteuteville , also I'm pretty baffled by a python file living under `fastsim-core/src/`

robinsteuteville: @calbaker see my earlier teams comment: "I also added a prepare_eff_values.py file to the same folder (wasn't sure where to add it), since I wanted to document how to transition the Altrios arrays to fastsim -- they needed to be transposed in a specific way to work given how the arrays in the interpolator need to be structured. That file can be moved or gotten rid of if desired."
I added this here more as a temporary measure, since I thought it was something that might need to be documented and I wasn't sure where it should be stored.

fastsim-core/src/vehicle/powertrain/efficiency_values.rs
robinsteuteville: @calbaker so to make sure I understand correctly, I should store the arrays as yaml files? I'm guessing I'll need to find something that deserializes the yaml file into a nested vector rather than a string?

robinsteuteville:
@calbaker here's my earlier teams message copied here:
here's the pr for the battery maps/corresponding methods. I added an example for how to use the methods to the demo_bev.py file, so you can play around with that if you want to convince yourself that it works as intended. I didn't add any tests on the rust side -- just confirmed on the python side that the methods worked, but I can add those tests if you'd like (the only rust test I can think of would be confirming that the res fields are updated as desired, but that can be easily checked by running the method in python as in the example I wrote). In order to store the default values, I created an efficiency_values.rs file with some constants stored, since the arrays get pretty big and I didn't want them gunking up the reversible_energy_storage.rs file. I also added a prepare_eff_values.py file to the same folder (wasn't sure where to add it), since I wanted to document how to transition the Altrios arrays to fastsim -- they needed to be transposed in a specific way to work given how the arrays in the interpolator need to be structured. That file can be moved or gotten rid of if desired. Finally, for some reason when I run sh build_and_test.sh I am getting the following failed test:
>       assert t_fsim2_mean / t_fsim3_no_save_mean > min_speed_ratio_si_none, \
            f"`min_speed_ratio_si_none`: {min_speed_ratio_si_none:.3G}, achieved ratio: {(t_fsim2_mean / t_fsim3_no_save_mean):.3G}"
E       AssertionError: `min_speed_ratio_si_none`: 4, achieved ratio: 3.8
E       assert (0.0006711803337869545 / 0.00017681933301597988) > 4.0

python/fastsim/tests/test_speedup.py:138: AssertionError
=================================================================================================== short test summary info ===================================================================================================
FAILED python/fastsim/tests/test_speedup.py::test_conv_speedup - AssertionError: `min_speed_ratio_si_none`: 4, achieved ratio: 3.8

so in short: the methods seem to work, and the altrios battery maps are added. However, I'm getting one failed test and I'm not sure why. The test doesn't seem to affect compilation or the ability of the methods to work on the python side

fastsim-core/src/vehicle/powertrain/efficiency_values.rs
calbaker: @robinsteuteville, ask Kyle what he thinks about this.  It could be the case that what I'm proposing is too onerous to be worth doing.

fastsim-core/src/vehicle/powertrain/prepare_eff_values.py
calbaker: @robinsteuteville, I'll need to give this some thought.

fastsim-core/src/vehicle/powertrain/efficiency_values.rs
robinsteuteville: @calbaker talked with Kyle, and he mentioned that the interpolator as a whole is serializable and deserializable and therefore each component also should be. However, I'm having a bit of trouble isolating the nested vector part to serialize and deserialize. One simple workaround could be to save in a yaml file an entire "default" interpolator for each dimension, rather than just saving separately each x, y, f_xy, etc. array. Then, each `set_default_xd_interp` method could access and deserialize one of the yaml files and set it to be the interpolator.

calbaker: > One simple workaround could be to save in a yaml file an entire "default" interpolator for each dimension,

Yes!  This!  @robinsteuteville

kylecarow:
@robinsteuteville To resolve merge conflicts, you probably just need to delete the `.subtrees` folder in this branch and re-merge the lastest `fastsim-3` changes. There may be some minor other changes to address but they shouldn't be difficult.

calbaker:
> @robinsteuteville To resolve merge conflicts, you probably just need to delete the `.subtrees` folder in this branch and re-merge the lastest `fastsim-3` changes. There may be some minor other changes to address but they shouldn't be difficult.

@robinsteuteville , FYI, you had deleted a single file in the `./.subtrees/` folder, but the whole folder should have been deleted, and merging `fastsim-3` into this branch took care of that, except for the conflict created by your deletion of the single file, which I've now fixed.  Merge conflicts usually should be resolved by pulling from the upstream branch (`fastsim-3`) rather than repeating the operation that was performed in that branch independently in your working branch.

calbaker:
@robinsteuteville , since I fixed the conflicts, is this now ready for review?

robinsteuteville:
> @robinsteuteville , since I fixed the conflicts, is this now ready for review?

@calbaker Last time we checked in about this, you were going to think some things through and perhaps change some things related to how the code was structured. If you've thought things through and feel it is good to go, then yes. There's nothing on my end that I still need to do (unless you find anything else you'd like me to change or update).

PR #170, opened 2024-12-02
creating to_pydict and from_pydict methods in fastsim-2

robinsteuteville:
[Corresponds to GitHub issue #356](https://github.nrel.gov/MBAP/fastsim/issues/356)

### DIFF
--- a/python/fastsim/__init__.py
+++ b/python/fastsim/__init__.py
@@ -72,11 +72,7 @@ def to_pydict(self) -> Dict:
     Returns self converted to pure python dictionary with no nested Rust objects
     """
     import json
-    try:
-        return json.loads(self.to_json())
-    except:
-        import yaml
-        return yaml.loads(self.to_yaml())
+    return json.loads(self.to_json())
 
 @classmethod
 def from_pydict(cls, pydict: Dict) -> Self:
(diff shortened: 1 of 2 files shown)


robinsteuteville:
@calbaker I added `from_pydict` and `to_pydict` methods and added them to the `RustVehicle` struct. A couple notes:
1. I have yet to figure out how to implement the methods for `fastsim.vehicle.Vehicle`, so they can only be used with either `to_rust` after the instantiation or using `fsr.RustVehicle` (see the demo file I created for a working example). The issue I ran into was that there was no method `from_json` or `to_json` for `fastsim.vehicle.Vehicle`.
2. I am happy to manually add code to implement the methods for additional Rust structs, however, if we want the methods to automatically be added to all FASTSim Rust structs, ideally we would use the automatically created `ACCEPTED_RUST_STRUCTS` list. This is currently used in FASTSim-3, however, when taking the same code and exploring with it to see if a similar code could work in FASTSim-2, I found that the FASTSim-3 code did not work -- I've confirmed that the same code does work in FASTSim-3, so I'll have to do some more exploring to figure out what needs to change for the FASTSim-2 version. So far I haven't had any luck figuring out what will work.
3. `test_fetch_github_list` is failing, but this may be a broader issue rather than branch specific.

calbaker:
@robinsteuteville , could you make this identical to the equivalent in fastsim-3?  No particular urgency.

calbaker:
@kylecarow @robinsteuteville  is this really even needed anymore?

calbaker (approved):

PR #172, opened 2024-12-06
PhantomData in structs to prevent outside instantiation

michael-okeefe:
Addresses:
https://github.nrel.gov/MBAP/fastsim/issues/362

Note: I ended up using the `#[non_exhaustive]` meta word syntax as it does prevent Struct Instantiation outside of the crate (e.g., I tested from fastsim-cli and the compiler would not let me struct instantiate).


calbaker (approved):
> Note: I ended up using the #[non_exhaustive] meta word syntax as it does prevent Struct Instantiation outside of the crate (e.g., I tested from fastsim-cli and the compiler would not let me struct instantiate).

@michael-okeefe I came here to ask this very question and you preempted me!    Nice work!

PR #180, opened 2025-01-14
Convenience methods for ninterp

kylecarow:
Solves https://github.nrel.gov/MBAP/fastsim/issues/382

### DIFF
--- a/fastsim-core/src/utils/interp.rs
+++ b/fastsim-core/src/utils/interp.rs
@@ -1,0 +1,1 @@
+use crate::imports::*;
--- a/fastsim-core/src/utils/interp.rs
+++ b/fastsim-core/src/utils/interp.rs
@@ -1,1 +1,107 @@
 use crate::imports::*;
+
+/// Methods for proportionally scaling interpolator function data
+pub trait InterpolatorMethods {
+    fn set_min(&mut self, min: f64) -> anyhow::Result<()>;
+    fn set_max(&mut self, max: f64) -> anyhow::Result<()>;
+    fn set_range(&mut self, range: f64) -> anyhow::Result<()>;
+}
+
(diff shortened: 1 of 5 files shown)


kylecarow:
@calbaker it ended up being much less of a headache just having the code directly in the `InterpND` match arm rather than fussing with dependencies

kylecarow:
@calbaker these should be ready to use, but maybe look over the `set_range` method. We might want to add [these checks](https://github.com/NREL/altrios/blob/981edf47bba724fb2d28f42dd6891637b7e81a2a/rust/altrios-core/src/macros.rs#L70-L79)

fastsim-core/src/utils/interp.rs
```diff
@@ -1 +1,107 @@
...
+            Interpolator::InterpND(interp) => {
+                todo!()
+            }
+        }
+    }
+
+    fn set_max(&mut self, max: f64) -> anyhow::Result<()> {
+        let old_max = self.max()?;
+        match self {
+            Interpolator::Interp0D(value) => Ok(*value = max),
```
kylecarow, line 33: @calbaker i thought this line was pretty wacky, mutating the value in a tuple variant

fastsim-core/src/utils/interp.rs
```diff
@@ -1 +1,107 @@
...
+                    })
+                    .collect(),
+            )?),
+            Interpolator::InterpND(interp) => {
+                Ok(interp.set_values(interp.values().map(|x| x * max / old_max))?)
+            }
+        }
+    }
+
+    fn set_range(&mut self, range: f64) -> anyhow::Result<()> {
```
kylecarow, line 61: @calbaker check the `ensure!`s and closures in this method

calbaker (approved):

PR #197, opened 2025-02-14
HEV transmission

kylecarow:
need to make sure the state updates and we use the efficiency etc.

### DIFF
--- a/cal_and_val/f3-vehicles/2016 FORD C-MAX HEV.yaml
+++ b/cal_and_val/f3-vehicles/2016 FORD C-MAX HEV.yaml
@@ -652,6 +652,11 @@ pt_type:
       specific_pwr_watts_per_kilogram: ~
       mass_kilograms: ~
       save_interval: 1
+    transmission:
+      mass_kilograms: ~
+      eff_interp:
+        Interp0D: 0.95
+      save_interval: 1
     pt_cntrl:
       RGWDB:
         speed_soc_disch_buffer: 17.8816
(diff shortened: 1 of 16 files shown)


kylecarow:
@calbaker any thoughts on how exactly to handle transmission efficiency in HEV's solve?

kylecarow:
@calbaker make sure to consider HEV thermal cal stuff before merging

calbaker:
> @calbaker make sure to consider HEV thermal cal stuff before merging

Is this otherwise ready to merge?

kylecarow:
@calbaker I think so, yes. Were the HEVs just not accounting for any transmission efficiency before? This PR adds a transmission efficiency penalty on the required power, feeding that into fc/em calculations.

calbaker:
@kylecarow , it looks like this needs to be handled in
https://github.com/NREL/fastsim/blob/efbc34cbd3a58ff48a9b04a0555f3ef405e8d54c/fastsim-core/src/vehicle/vehicle_model/fastsim2_interface.rs#L99

Could you take a look and implement if you agree?

kylecarow:
> Could you take a look and implement if you agree?

@calbaker Is that addressed here?

https://github.com/NREL/fastsim/blob/efbc34cbd3a58ff48a9b04a0555f3ef405e8d54c/fastsim-core/src/vehicle/vehicle_model/fastsim2_interface.rs#L201-L207

calbaker:
> > Could you take a look and implement if you agree?
> 
> @calbaker Is that addressed here?
> 
> https://github.com/NREL/fastsim/blob/efbc34cbd3a58ff48a9b04a0555f3ef405e8d54c/fastsim-core/src/vehicle/vehicle_model/fastsim2_interface.rs#L201-L207

yep!

calbaker (approved):

PR #205, opened 2025-04-09
Fix/feature gate numpy

calbaker:


kylecarow:
@calbaker can we just remove this dep?

calbaker:
> @calbaker can we just remove this dep?

Are we not using it?  I didn't really dive into that.

PR #207, opened 2025-04-11
adding `TrackedState` struct to verify mutation behavior

calbaker:
This PR is a significant change with many improvements:
- `TrackedState<T>` struct to wrap all state variables to ensure mutation once and only once per time step.  These checks result in modest performance losses but are deemed highly valuable in spite of that!
-  separated out serde and pyo3 api stuff
    ```rust
    #[serde_api] // this must always be the first proc macro to invoke for the derive macros to work
    #[derive(Deserialize, Serialize, Debug, Clone, PartialEq, StateMethods, SetCumulative)]
    /// Struct for modeling [FuelConverter] (e.g. engine, fuel cell.) thermal plant
    #[non_exhaustive]
    #[serde(deny_unknown_fields)]
    #[cfg_attr(feature = "pyo3", pyclass(module = "fastsim", subclass, eq))]
    pub struct FuelConverter {
    ```
- pyo3 api is now placed above impl block
    ```rust
    #[named_struct_pyo3_api]
    impl FuelConverter {
        // optional, custom, struct-specific pymethods
        #[getter("eff_max")]
        fn get_eff_max_py(&self) -> PyResult<f64> {
            Ok(self.get_eff_max()?)
        }
    ```


fastsim-core/src/utils/tracked_state.rs
calbaker, line 1: @michael-okeefe and @kylecarow, please review this file carefully!

fastsim-core/src/utils/tracked_state.rs
```diff
@@ -0,0 +1,240 @@
...
+    {
+        let value: T = T::deserialize(deserializer)?;
+
+        Ok(Self(value, Default::default()))
+    }
+}
+
+#[cfg(test)]
+mod test_tracked_state {
+    // use super::*;
```
michael-okeefe: Should all of the test code be commented out? Seems like this would be handy to have to ensure the implementation is working correctly.

michael-okeefe (commented):
I've gone through all the code visually. I believe I'm following what you are trying to do and it appears correct but there is certainly a lot here -- including macros which are notoriously difficult to tell if they're doing what is intended with only a visual inspection. I think my only general comment would be to ensure you have enough tests / testing coverage to give you confidence that things appear to be in order (you may have that already).

fastsim-core/src/utils/tracked_state.rs
```diff
@@ -173,68 +173,40 @@ where
...
-    //             *pwr.get_fresh(|| format_dbg!()).unwrap() * *dt.get_fresh(|| format_dbg!()).unwrap(),
-    //             format_dbg!(),
-    //         )
-    //         .unwrap();
-
-    //     pwr.is_fresh_and_reset(format_dbg!()).unwrap();
-    //     dt.is_fresh_and_reset(format_dbg!()).unwrap();
-    //     energy.is_fresh_and_reset(format_dbg!()).unwrap();
-    // }
+    use super::*;
```
calbaker, line 176: @michael-okeefe , if these tests look appropriate to you, please go ahead and approve and merge unless there are other concerns you might have.

fastsim-core/src/utils/tracked_state.rs
```diff
@@ -0,0 +1,214 @@
...
+        D: serde::Deserializer<'de>,
+    {
+        let value: T = T::deserialize(deserializer)?;
+
+        Ok(Self(value, Default::default()))
+    }
+}
+
+#[cfg(test)]
+mod test_tracked_state {
```
calbaker, line 175: @michael-okeefe , if these tests look appropriate to you, please go ahead and approve and merge unless there are other concerns you might have.

michael-okeefe (approved):

PR #211, opened 2025-04-25
remove unnecessary argmin ndarray dep

kylecarow:
Removes unnecessary argmin ndarray dep, replacing with vectors internally

Stops us from having multiple ndarray versions, and just supports 0.16.x

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -17,7 +17,7 @@ codegen-units = 1 # optimize connection between modules
 
 [workspace.dependencies]
 anyhow = "1.0.57"
-ndarray = { version = ">=0.15.3, <0.17", features = ["serde"] }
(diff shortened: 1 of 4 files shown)


rust/Cargo.toml
```diff
@@ -17,7 +17,7 @@ codegen-units = 1 # optimize connection between modules
 
 [workspace.dependencies]
 anyhow = "1.0.57"
-ndarray = { version = ">=0.15.3, <0.17", features = ["serde"] }
+ndarray = { version = "0.16.1", features = ["serde"] }
```
calbaker, line 20: @kylecarow , I'm going to approve this, but keep in mind that having a firmly nailed down version requirement could lead to trouble later on.

calbaker (approved):

PR #218, opened 2025-05-02
Update release.yaml and build_and_test.yaml with fixes.

michael-okeefe:
Updating GitHub workflows to incorporate the latest materials. This PR is for the fastsim-3 branch (and release v3.0.0). When we get this working, we'll also make a PR for the fastsim-2 branch.

Some of the issues being addressed are mainly around updating the build scripts to work with Apple Intel and Apple Silicon.

### DIFF
--- a/.github/workflows/build_and_test.yaml
+++ b/.github/workflows/build_and_test.yaml
@@ -42,9 +42,9 @@ jobs:
       - name: set up rust
         uses: dtolnay/rust-toolchain@stable
 
-      - name: install rust target for MacOS 14 (Apple Silicon)
+      - name: install rust target for MacOS 14+ (Apple Silicon)
         run: rustup target add aarch64-apple-darwin x86_64-apple-darwin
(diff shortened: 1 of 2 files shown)


michael-okeefe:
Note: I also triggered the `release.yaml` workflow which now has a "deploy guard". That is, on manual trigger, the first part of the file will run but the second part (which deploys to PyPI) will not. This will allow us to confirm that the file works up to that point at least...

You can watch it here:

https://github.com/NREL/fastsim/actions/runs/14797636776

All tests and the first part of the release workflow should all run green prior to moving forward with another release attempt.

michael-okeefe:
OK, looks like the first part of `release.yaml` works perfectly:

https://github.com/NREL/fastsim/actions/runs/14797636776

The deployment part is the only piece not fully tested. We'll have to test when we try release again next...

calbaker (approved):

PR #219, opened 2025-05-02
Update release.yaml to match latest from fastsim-3.

michael-okeefe:
The only change from what's on fastsim-3 is in
the call for `cargo test` as we need to cd into
the rust directory on fastsim-2. Otherwise, they
are the same.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -5,6 +5,8 @@ on:
     types: [published]
   workflow_dispatch:
 
+# NOTE: parts inspired by https://cibuildwheel.pypa.io/en/stable/setup/
+
 jobs:
   build:
     name: build py3.${{ matrix.python-version }} on ${{ matrix.platform || matrix.os }}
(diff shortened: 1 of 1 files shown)


michael-okeefe:
Currently running `release.yaml` to ensure it is working correctly. The deploy step has guards in it to prevent deployment to PyPI but the first part of the file will run to assure us it is actually working (which is the point of this PR).

When we get all green lights, I'll move this PR from draft to final:

https://github.com/NREL/fastsim/actions/runs/14801972411

.github/workflows/release.yaml
```diff
@@ -5,76 +5,81 @@ on:
...
-          - ubuntu
-          - macos
-          - windows
+          - ubuntu-latest
+          - macos-13 # Intel runner
+          - macos-latest # Apple Silicon runner
+          - windows-latest
         python-version:
-          - "9"
           - "10"
```
calbaker, line 22: @michael-okeefe, do you know if there's a  way to have it obtain the python versions from the pyproject.toml file.  That would be pretty slick.  I'll wait for your response before merging.

michael-okeefe: @calbaker 

I did some quick searching and came across this:

https://github.com/BerriAI/litellm/blob/main/.github/workflows/read_pyproject_version.yml

So it is possible to call Python itself to dig data out of the pyproject.toml -- one question is whether than can be called to dynamically set the run matrix or not... and the answer appears to be YES:

https://github.com/pypa/cibuildwheel/discussions/1261

So it definitely looks possible. If this looks like something worth pursuing, I can take a crack at it starting on Wednesday this week. We'd have to give it a shot and trial run the Github Action to actually confirm it is working.

calbaker (approved):

calbaker:
@michael-okeefe , I attempted to resolve a merge conflict in 602c5993.  Could you check this?

michael-okeefe:
> @michael-okeefe , I attempted to resolve a merge conflict in [602c599](https://github.com/NREL/fastsim/commit/602c59936c961ba747714b0edca1959e9d3cb135). Could you check this?

Looks good, @calbaker . Thank you for addressing the merge conflict.

PR #194, opened 2025-02-13
Completing TODOs where able

robinsteuteville:

PR #229, opened 2025-05-22
F3/pyo3 api cleanup

calbaker:


fastsim-core/src/drive_cycle.rs
```diff
@@ -588,4 +582,17 @@ mod tests {
                 .collect::<Vec<si::Length>>()
         );
     }
+
+    type StructWithResources = Cycle;
+
+    #[test]
+    fn test_resources() {
```
kylecarow, line 589: @calbaker does this just cover drive cycles? I like this idea, and asked @robinsteuteville to make a script to do something similar in #194

fastsim-core/src/traits/serde_api.rs
```diff
@@ -0,0 +1,364 @@
...
+        let file = Self::RESOURCES_DIR.get_file(&filepath).ok_or_else(|| {
+            Error::SerdeError(format!("File not found in resources: {filepath:?}"))
+        })?;
+        Self::from_reader(&mut file.contents(), extension, skip_init)
+    }
+
+    /// List the available resources in the resources directory
+    ///
+    /// RETURNS: a vector of strings for resources that can be loaded
+    fn list_resources() -> Result<Vec<PathBuf>, Error> {
```
kylecarow, line 52:
```suggestion
    #[cfg(feature = "resources")]
    fn list_resources() -> Result<Vec<PathBuf>, Error> {
```

fastsim-core/src/drive_cycle.rs
calbaker: @kylecarow , yep!

PR #243, opened 2025-12-02
Non-serde constructors

kylecarow:
Making a draft PR for this so it doesn't slip off my radar


kylecarow:
note to self and @robinsteuteville: Create constructors for higher-level vehicle structs and anything else that makes sense

kylecarow:
Let's suggest using pydicts instead of introducing a bunch of new API methods for the current version of FASTSim. I'll leave this branch non-deleted for future reference though.

PR #253, opened 2026-04-02
F3/add start stop

michael-okeefe:
This PR adds a stop/start functionality to FASTSim 3 for both the conventional and hybrid powertrains (i.e., mild- or micro- hybrid). The functionality is implemented by adding two new control options (one for Conventional and one for Hybrid Electric Vehicles) called StopStart.

In addition, this PR also adds a python demo file (`demo_stop_start.py`) which showcases how to use the new functionality.

Reference:
- internal MBAP issue 418


michael-okeefe:
Note: updated to better configure the micro-hybrid case. Also, tweaked pandas version to prevent build issue on python 3.10 for pandas (same fix that was applied on f3/add-dfco branch).

PR #254, opened 2026-04-14
F3/add dfco

michael-okeefe:
Add decel fuel cut-off (DFCO) capabilities to FASTSim 3.

A demo file, `demo_dfco.py` file is added to show how to configure and use the DFCO system.

Internal issue reference: https://github.nrel.gov/MBAP/fastsim/issues/417


michael-okeefe:
Added fix to prevent build issue on Linux for python 3.10 with pandas. Prevented using pandas 2.3.3 and only allow up to 2.3.2.

michael-okeefe:
Note: this PR contains all the work for DFCO as well as stop/start ( #253 ).

jakeholden:
Thanks, @michael-okeefe! I just looked through all of this and it looks excellent! DFCO and start/stop savings are pretty well aligned with what we expected. A couple of quick notes before before I approve/merge just to make sure I understand everything: 

## PR Review Notes

### 1. DFCO Demo Label Fix (minor — already applied)
The labels in `demo_dfco.py` had leftover references to `"ss"` (from the stop/start demo). I updated them to `"dfco"` to match the actual feature being demonstrated.

### 2. Pacifica Vehicle Model — Missing `pt_cntrl` Parameter
I noticed that many of the vehicles in `cal_and_val` were updated to include the `pt_cntrl` parameter. Should the Pacifica model (`2026_Chrysler_Pacifica_Select.yaml`) also be updated to include that parameter?

michael-okeefe:
Thank you, @jakeholden .

> ### 1. DFCO Demo Label Fix (minor — already applied)
> 

Thank you -- good catch. I just pushed that change over from the internal github to the public one for this PR as well.

> ### 2. Pacifica Vehicle Model — Missing `pt_cntrl` Parameter
> 
> I noticed that many of the vehicles in `cal_and_val` were updated to include the `pt_cntrl` parameter. Should the Pacifica model (`2026_Chrysler_Pacifica_Select.yaml`) also be updated to include that parameter?

The vehicles in `cal_and_val` got automatically updated for me by virtue of running the Rust unit tests. For the Pacifica, it won't hurt anything to not have the parameter -- it would just load up with defaults. However, we could certainly add it and we definitely should add it if we would like to have the Pacifica default to using stop/start (though we'd have to slightly rework the demo example). What is your preference?

jakeholden:
> The vehicles in `cal_and_val` got automatically updated for me by virtue of running the Rust unit tests. For the Pacifica, it won't hurt anything to not have the parameter -- it would just load up with defaults. However, we could certainly add it and we definitely should add it if we would like to have the Pacifica default to using stop/start (though we'd have to slightly rework the demo example). What is your preference?

@michael-okeefe this was my suspicion, thanks for confirming! I don't think we need to set it to using stop/start for the sake of this PR, though we may change that eventually. I just wanted to make sure my understanding was correct. I'll go ahead and approve and merge!

jakeholden (approved):
Excellent work, @michael-okeefe! Fuel savings from start/stop and DFCO demos on UDDS both result in the right approximate savings that we were expecting.

michael-okeefe:
> Excellent work, @michael-okeefe! Fuel savings from start/stop and DFCO demos on UDDS both result in the right approximate savings that we were expecting.

That's great to hear! Thank you.

PR #255, opened 2026-04-24
feat: added direnv as a convenience

calbaker:
automatically runs `uv sync` and activates python venv whenever:
- user uses `cd` to get to the project dir
- user runs `git checkout`


kylecarow:
Guten Tag Chad! Apologies it took me so long to look at this PR. Hope all is going well :)

I've added a few lines to handle pixi vs. uv+venv a bit nicer, prioritizing pixi if it is found. The `.envrc` file also activates the `dev` pixi environment by default, which I think is the sensible choice for someone who is using FASTSim + pixi + direnv.
I also added rust as a pixi dev dependency which is a little unrelated to this PR.

I will go ahead and merge, but let me know if you have any thoughts or qualms about my changes.

PR #260, opened 2026-06-23
fixed endless SOC iteration for HEVs

kylecarow:
@jhoshiko @jakeholden 

I've fixed the issue with endless SOC iteration that we were seeing in https://github.nrel.gov/MBAP/fastsim/issues/423.

The issue was that when starting a new iteration, only the ending SOC -> starting SOC carryover was plumbed up correctly. The actual iteration counter was getting reset each iteration as it is stored within the Vehicle object.

This PR does **_not_** fix the issues with those specific vehicles not converging to a balanced SOC!


kylecarow:
Worth noting too that the default maximum iterations in fastsim-3 is `5`, which is decreased from the fastsim-2 maximum of `30`(!!!)

@jakeholden Should we change this limit back?

kylecarow:
@jhoshiko the error for > 5 iterations seems like its pretty vague, let me know if the message you see should be changed :)

https://github.com/NatLabRockies/fastsim/blob/e05cacf41a874edfa7810b03aeccac481dd7df19/fastsim-core/src/simdrive/mod.rs#L176-L204

kylecarow:
@jhoshiko can this be merged?

jhoshiko (approved):
I'd Say So!

PR #264, opened 2026-06-29
Use Pixi for CI/CD setup

kylecarow:
Affects `deploy_docs` and `build_and_test`, not `release` (yet)

Changes to `pyproject.toml`
- Reorganize optional-dependencies into `dependency-groups`, this aligns better with the philosophy here
https://pydevtools.com/handbook/explanation/what-are-optional-dependencies-and-dependency-groups/
  - no end users should be using `pip install '.[dev]'`/`pip install fastsim[dev]` so I'm not fussed about compatibility changes. You can still install these deps by doing `pip install --group dev .` or `pip install --group dev -e .` anyway.
  - I left a `pymoo` optional-dependency in there commented out, it seems like that would be a sensible optional dependency. The way we reference pymoo in this repo is messy anyway and should generally be cleaned up.
- `project.requires-python` bumped to allow Python 3.10 through 3.14
- New Pixi environments to facilitate Python version matrix

Workflows use pixi for Rust setup:
- Each currently supported Python version gets its own pixi env with specified dependency
- Reorganized `deploy_docs` and `build_and_test` workflows to use pixi for Rust installation, Python installation, FASTSim, and extra dependencies. docs workflow uses new `docs` pixi env that does not include any extra deps beyond what is required.
- Is Rust caching step necessary? The robots thought that line ought to go, but perhaps it's still useful?

### DIFF
--- a/.github/workflows/build_and_test.yaml
+++ b/.github/workflows/build_and_test.yaml
@@ -26,9 +26,12 @@ jobs:
             version: latest
             arch: "amd64"
         python-version:
+          # https://devguide.python.org/versions/#status-of-python-versions
           - "10"
           - "11"
           - "12"
+          - "13"
+          - "14"
         include:
           - os:
               base: ubuntu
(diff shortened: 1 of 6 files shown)


kylecarow:
Superseded by #265

PR #267, opened 2026-07-09
F3/docs/demos

kylecarow:


kylecarow:
@jhoshiko When you get a chance, and this is probably low priority, could i get your eyes on https://github.com/NatLabRockies/fastsim/blob/f3/docs/demos/docs/content/developers/making-releases.md and let me know what you think of my rules for semver in FASTSim?

Essentially if formalizes how we version FASTSim and when we introduce breaking changes.

@robinsteuteville and @jakeholden it's probably good for you to read these over as well so we can consider these when adding features, changing fields etc

robinsteuteville:
Comments on the docs:
Looks great to me! Have a few minor thoughts:

FASTSim Documentation:
- Wondering if the definition we give for FASTSim at the top of the page is too narrow? it models technology improvement effects, but also current/historical vehicles for a wide range of cycles, which I feel like is an important function in its own right.

What is FASTSim
- The definition at the top of the page could be a little clearer and more specifiv. Perhaps explicitly saying it models vehicle powertrains rather than the more vague "answers common technology-assessment questions" -- I feel like the description of fastsim at the beginning of "Getting Started" is a good place to start

Getting Started
- at the top, it says it is an "interactive demonstration of FASTSim" -- not sure if this is accurate/needed anymore?

What is a SimDrive Object?
- for the highway/city cycle example simulating the same vehicle over both cycles, I only see code for the highway cycle, not the city cycle

Editing Simulation Parameters
- Getting a runtime error for cell 2: "state variable has not been updated. this is a bug in 'fastsim-core'.

Running Thermal Simulations
- Wondering if there needs to be any additional background before getting into the comparisons? Might be good to explain the structure of thermal modeling inputs within fastsim and what additional inputs are needed or can be added for a thermal model. Maybe even briefly mention how to turn a non-thermal model into a thermal model.

kylecarow:
@robinsteuteville Do these changes address your concerns?

FASTSim Documentation:  
> Wondering if the definition we give for FASTSim at the top of the page is too narrow? it models technology improvement effects, but also current/historical vehicles for a wide range of cycles, which I feel like is an important function in its own right.  

Replaced with the following:
> FASTSim (**Future Automotive Systems Technology Simulator**) is a vehicle simulation tool for evaluating the efficiency and performance of on-road vehicles. Developed by the National Laboratory of the Rockies (NLR), it provides fast, accurate estimates of powertrain performance and energy consumption. FASTSim models powertrain technologies from conventional fuel-burning vehicles to hybrids and battery electric vehicles, and has been used to simulate light-duty passenger cars, two-wheelers, and a wide variety of medium- and heavy-duty vocational vehicles.

Inspired by the getting started demo paragraph and broadened the scope a bit.

---

What is FASTSim  
> The definition at the top of the page could be a little clearer and more specifiv. Perhaps explicitly saying it models vehicle powertrains rather than the more vague "answers common technology-assessment questions" -- I feel like the description of fastsim at the beginning of "Getting Started" is a good place to start  

Replaced with the paragraph above, as well as these lines:

> FASTSim is used to answer common technology-assessment questions quickly, with fewer required inputs than high-fidelity physics models. This page describes FASTSim's history and the level of fidelity to real-world performance it aims to achieve.

---

Getting Started  
> at the top, it says it is an "interactive demonstration of FASTSim" -- not sure if this is accurate/needed anymore?  

Getting Started paragraph now reads:

> This is a demonstration and high-level overview of FASTSim. Developed by the National
Laboratory of the Rockies (NLR), FASTSim provides extremely fast and accurate
estimates of powertrain performance and energy (fuel or electricity) consumption for
a wide variety of vehicle types. This page walks through the core workflow:
loading a vehicle, loading a drive cycle, running a simulation, and working
with the results.

---

What is a SimDrive Object?  
> for the highway/city cycle example simulating the same vehicle over both cycles, I only see code for the highway cycle, not the city cycle  

Yeah the fact this was reusing earlier cell output is confusing. I changed it to explicitly re-run the city simulation side-by-side.

---

Editing Simulation Parameters  
> Getting a runtime error for cell 2: "state variable has not been updated. this is a bug in 'fastsim-core'.  

I think you must be running on a compiled FASTSim from another branch, that page runs without error for me, and `pytest` passes both locally and on CI.

---

Running Thermal Simulations  
> Wondering if there needs to be any additional background before getting into the comparisons? Might be good to explain the structure of thermal modeling inputs within fastsim and what additional inputs are needed or can be added for a thermal model. Maybe even briefly mention how to turn a non-thermal model into a thermal model.

Added a background section to the page:

> FASTSim has optional vehicle model components for thermal simulation:
> - `cabin`: Models cabin temperature and heat transfer
> - `hvac`: Models HVAC performance
> - FC `thrml`: Models fuel converter (e.g. internal combustion engine) thermal performance
> - RES `thrml`: Models reversible energy storage (i.e. traction battery) thermal performance
> 
> See [Vehicles in FASTSim]() for a complete description of each thermal component.

Issue #270, opened 2026-07-12
Tracking issue for fuel cell electric vehicles


kylecarow:
Tracking issue to list notes/problems that need to be fixed before fuel cell vehicles will work in FASTSim-3

<img width="1084" height="644" alt="Image" src="https://github.com/user-attachments/assets/07d3b840-4ae3-423a-95a8-74f531260a47" />

- ~Internally we will handle a fuel cell vehicle as a HybridElectricVehicle, but add a new `PowertrainType` option `FuelCellElectricVehicle(Box<HybridElectricVehicle>)`~
- New powertrain type and inner struct:
  ```rust
  pub enum PowertrainType {
      // ...
      #[serde(rename = "FCEV")]
      #[serde(alias = "FuelCellElectricVehicle")]
      FuelCellElectricVehicle(Box<FuelCellElectricVehicle>),
  }
  ```
- Copy series HEV logic into new `FuelCellElectricVehicle` inner struct
- Why not reuse `HybridElectricVehicle`?
  - No generator for FCEVs, FuelConverter supplies electrical power
    - No need to allow None generator for Series hybrids for just the FCEV case
  - HEVPowertrainControls options do not apply to FCEV
  - HEVAuxControls may still apply, dunno yet

FuelType added by PR #259 (issue #289) will be useful for this, it already has a `Hydrogen` variant

Blockers:
- #283
  - almost same topology as series hybrid, but FC supplies electrical power directly to RES/EM/aux(?) instead of through a generator
  - adding a series hybrid will teach a lot about FASTSims inner workings and is the same core logic
- #306 
- #286
- After these, create a new struct `FuelCellElectricVehicle`

PR #272, opened 2026-07-13
Alternator efficiency for HEVs (applies fc aux load only)

kylecarow:
This PR fixes a small bug I found where HEVs/PHEVs did not account for alternator efficiency in aux power loads coming supplied by the fuel converter. Aux loads supplied by the RES are assumed to be all electric as is our assumption for BEVs so there is no alternator to account for there.

In fact, we did not even have an alternator efficiency in the HybridElectricVehicle at all, so this PR adds that and makes a default of 100% to maintain compatibility with existing vehicle files (that would otherwise have a missing field and fail to deserialize). There is also a test to make sure this works as expected. Maybe we can add a logging event letting the user know the default value of 100% was used? I anticipate getting rid of this serde default as soon as we switch to a break in the vehicle file format (next major version?)

A hiccup I predict with this change: if we want to model fuel cells in FASTSim 3 as HEVs (but have a different powertrain type, just like we do for PHEVs), then having an alternator efficiency in the HybridElectricVehicle doesn't exactly make sense, as they have no alternator. We can enforce this field to be 100% but I really don't like the idea of having fields that don't apply - that feels like a FASTSim-2-ism.
- We could just make the alt_eff for HEVs `Option<si::Ratio>` instead, and make None unwrap to 100%
- Relevant #270

I also dont like how we model alternator as a ratio rather than an interpolator. Could we support alternators that have varying efficiency based on power load? FC temperature as a surrogate?

### DIFF
--- a/cal_and_val/f3-vehicles/2010 Mazda 3 i-Stop.yaml
+++ b/cal_and_val/f3-vehicles/2010 Mazda 3 i-Stop.yaml
@@ -152,6 +152,7 @@ state:
   i: 0
   time_seconds: 0.0
   pwr_prop_fwd_max_watts: 0.0
+  speed_trac_fwd_max_meters_per_second: 0.0
   pwr_prop_bwd_max_watts: 0.0
   pwr_tractive_watts: 0.0
   pwr_tractive_for_cyc_watts: 0.0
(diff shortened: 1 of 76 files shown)


kylecarow:
Should probably rebase these changes on #279 and implement this as a vehicle-wide aux load efficiency penalty

kylecarow:
#286 is a better approach

PR #259, opened 2026-06-18
Updates to allow for upgraded vehicle repository

kylecarow:
Resolves  #274 and #289

Master issue for all vehicle database work: #288

Based off of #265 / #267, review those first to not get overwhelmed by the huge diff

### New: Vehicle Database Schema v1

A new `DatabaseSchemaV1` struct (schema_v1.rs) formalizes the directory layout for the `fastsim-vehicles` repository. It encodes an 8-segment path:

```
v1/fastsim-{N}/{powertrain}/{make}/{model}/{year}/{variant}/v{N}
```

| Level | Description | Example |
| --- | --- | --- |
| Schema version | Describes the format of subsequent organizational levels. | `v1` |
| FASTSim version | Major FASTSim version. | `fastsim-3` |
| Powertrain | Powertrain type (e.g. `conv`/`hev`/`phev`/`bev`/`fcev`) | `hev` |
| Make | Vehicle make/manufacturer name. | `toyota` |
| Model | Vehicle model name and any applicable trim information. | `camry-xle` |
| Year | Model year or year range. | `2015` |
| Variant | Description of active modeling feature set. | `thermal` |
| Revision | Model revision number for updates and corrections. | `v1` |

Example full filepath:

`v1/fastsim-3/hev/toyota/camry-xle/2015/thermal/v1.yaml`

The struct serializes to/from a plain string, so it stores compactly in YAML as a single field.

---

### New: `Vehicle.from_db` and underlying Rust methods

The high-level Python entry point is `Vehicle.from_db(db_path_or_url, schema=1, **kwargs)`. `db_path_or_url` is `None` (default remote), an `https://...` URL, or a local filesystem path. It supports two input modes:

**Mode 1 — Path string** (caller supplies the pre-serialized schema path):
```python
Vehicle.from_db(
    db_path_or_url="/path/to/fastsim-vehicles",
    path="v1/fastsim-3/conv/ford/fusion/2012/base/v1",
    # optional: extension="yaml" (default), skip_init=False
)
# Extension can also be embedded directly in the path:
Vehicle.from_db(path="v1/fastsim-3/conv/ford/fusion/2012/base/v1.yaml")
# (cannot specify extension both in path and as a separate kwarg)
```

**Mode 2 — Individual fields** (schema path is constructed internally):
```python
Vehicle.from_db(
    db_path_or_url="/path/to/fastsim-vehicles",
    powertrain="conv",       # required: "conv", "hev", "phev", "bev"
    make="Ford",             # required
[...]

PR #291, opened 2026-07-16
Skip serializing history if it is empty

### LINKED ISSUE #274: Skip serializing `history` if it is empty
PR #259

kylecarow:
Also applies cargo fmt and regenerates cal_and_val vehicles

Resolves #274


kylecarow:
Absorbed into #259

PR #301, opened 2026-07-18
Organize and feature-gate FASTSim-2 compatibility code

kylecarow:
- [x] create `compat` mod and feature
  - [x] feature gate dependencies that are needed with updated Cargo.lock
    - something about running `cargo update` in fastsim-3 with the existing setup causes failure on some platforms, this is the fix (rather than never updating Cargo.lock)
- [x] move all compatibility code to `compat`
  - `from_f2_file` still needs to live in the impl with other pymethods, not a huge deal
  - [x] remove all remaining references to `fastsim2` in fastsim-3 outside of `compat::fastsim2`
- [x] `compat/assets` folder to organize any f2 assets (e.g. vehicle files)
  - [x] include_dir instead of include_str
  - [x] move `f2-*.yaml` loose files to `compat/assets`
- [x] refactor code that previously loaded fastsim-2 models to convert to fastsim-3 (e.g. mock_veh functions)
  - these now just load from resources
- [x] remove cal_and_val linkage to simdrivelabel
  - PHEV file just lives in fastsim-core now which makes more sense
- [x] add explicit test coverage for f2 to f3 vehicle conversion

# This allows us to load F2 files in F3 with no added fuss or functions!


fastsim-core/src/traits/serde_api.rs
```diff
@@ -215,8 +215,7 @@ pub trait SerdeAPI: Serialize + for<'a> Deserialize<'a> + Init {
                     let mut buf = String::new();
                     rdr.read_to_string(&mut buf)
                         .map_err(|err| Error::SerdeError(format!("{err}")))?;
-                    Self::from_toml(buf, skip_init)
-                        .map_err(|err| Error::SerdeError(format!("{err}")))?
+                    Self::from_toml(buf, true).map_err(|err| Error::SerdeError(format!("{err}")))?
```
kylecarow, line 230: this is an unrelated bug, skip init `true` prevents init from running twice in this call chain

fastsim-core/src/vehicle/vehicle_model.rs
```diff
@@ -371,7 +371,126 @@ impl Mass for Vehicle {
 impl SerdeAPI for Vehicle {
     #[cfg(feature = "resources")]
     const RESOURCES_SUBDIR: &'static str = "vehicles";
+
+    // specialized from_reader that allows for compatibility with fastsim-2 vehicle files
```
kylecarow: the specialized from_reader and from_* str methods are where the autoconversion magic happens, it takes a stab at deserializing from the FASTSim-2 format first, then falls back to normal deserialization. some finagling had to be done in from_reader to scroll back the buffer (otherwise it gets consumed in the compat pass)

fastsim-core/src/vehicle/vehicle_model.rs
```diff
@@ -1045,6 +1164,44 @@ pub(crate) mod tests {
 
     type StructWithResources = Vehicle;
 
+    #[test]
+    #[cfg(all(feature = "compat", feature = "yaml"))]
```
kylecarow, line 1438: added a couple tests that check from_reader and from_yaml work directly on the compat fastsim-2 vehicle asset files, and they do!

Issue #287, opened 2026-07-15
Efficiency modeling refinements master issue


kylecarow:
- #279
  - Follow-on work:
    - [ ] refactor to introduce `FCEfficiency`
    - [ ] refactor to introduce `EMEfficiency`
      - how do we handle EM having multiple interpolators?
    - [x] refactor to introduce `AuxSupplyEfficiency`
      - Issue #286
      - PR #304 
- Convenience functions for maps #275

PR #304, opened 2026-07-20
Introduce aux efficiency field `aux_eff` for FC and RES, deprecate Conv `alt_eff`

### LINKED ISSUE #286: Refactor `alt_eff` to be a FuelConverter/ReversibleEnergyStorage level aux load efficiency penalty
~Necessary for #270~ Not necessary, but useful

### Problem with current state
- In BEV, HEV, PHEV:
  - We do not account for electrical-electrical conversion (DC:DC) losses from the RES to aux power
- In HEV:
  - We do not account for mechanical-electrical conversion (alternator) losses from the FC to aux power
    - `FuelConverter` has `alt_eff` that accounts for this
- For FCEVs (modeled as a powertrain type around `HybridElectricVehicle`): there should be no alternator, so how should this be accounted for?

- ~Potential solutions, but not ideal:~
  - ~Add alt_eff to `HybridElectricVehicle` (as in #272)~
    - ~Does not apply when we get to modeling FCEVs, FCEVs have no alternator~
  - ~Introduce a single vehicle-level field~
    - ~can't distinguish between mechanical (FC) vs electrical (RES) conversion losses~
  - ~Introduce multiple vehicle-level fields, for FC and RES~
    - ~possibility of modeling inapplicable vehicles (conventional vehicle with a RES aux load efficiency?)~

### Solution: Component-Level Efficiency Fields

**Add optional `aux_load_eff: Option<si::Ratio>` to:**
- `FuelConverter` — represents mechanical-to-electrical conversion (alternator for ICE, inapplicable for fuel cell)
- `ReversibleEnergyStorage` — represents electrical-to-electrical conversion (DC-DC inverter)

### Add initialization checks
- **Conv** init: 
  - FC: (0, 100%] range check on aux_load_eff
- **HEV/PHEV** init: 
  - FC: (0, 100%] range check on aux_load_eff
  - RES: (0, 100%] range check on aux_load_eff
- **FCEV** init:
  - FC: `ensure!(fc.aux_load_eff == None or Some(1.0))`
  - RES: (0, 100%] range check
  -FC does not supply aux load directly for FCEVs, it comes from the RES

- [x] create new ~optional~ field in FC and RES components for aux load efficiency penalty
  - FC: represents an alternator (fuel cell stacks dont directly supply aux power, all goes to RES)
    - for FCEV, check on initialization this value is None or Some(1.0), e.g.:
```rust
impl Init for FuelCellElectricVehicle {
    fn init(&mut self) -> Result<(), Error> {
        // Validate that FC doesn't have alternator (FCEV is purely electrical)
        if let Some(eff) = self.fc.aux_load_eff {
            ensure!(
                eff == 1.0 * uc::R,
                "FCEV fuel converter should not have alternator losses. aux_load_eff must be None or 1.0, got {}",
                eff.get::<si::ratio>()
            );
        }
        
        // ... rest of init
        self.fc.init()?;
        self.res.init()?;
        // ...
        Ok(())
    }
}
```
  - RES: represents DC:DC conversion efficiency
- [x] figure out how to maintain serde format compatibilty

~Eventually port to style of #279, delegate some efficiency range checks to that implementation~ Did this already, why not

kylecarow:
Resolves #286

Build upon PR #279 (thus related to #287). PR starts at [a3b0320](https://github.com/NatLabRockies/fastsim/pull/304/commits/a3b032039ff4c6fecaa54158b731207cbcf4b9c7)

This PR:
- Introduces new Option<AuxEfficiency> fields for the FC and RES (where aux load actually comes from) rather than using `alt_eff` at the powertrain level (which introduces problems for
  - this AuxEfficiency comes with all the standard efficiency range and value checks that PR #279 introduces
  - slightly refined the macro from that PR to add a From<f64> impl to easily create constant-efficiency instances
     - could also and a From<Ratio<64>> method as well
- plumbs up FC `aux_eff` to replace `alt_eff` for conventional vehicles
- plumbs up RES `aux_eff` to account for aux load efficiency from batteries (e.g. DC:DC conversion, etc.)
- deprecates `alt_eff`
  - allows it to still be read, parsed into aux_eff if its not set, and emitting appropriate warnings
  - no longer serialized, defaults to 1.0 (100%) when not provided (i.e. for newly serialized vehicles)

Modeling improvements:
- All conventional vehicles no longer require an alternator efficiency, instead it is handled by the optional FC aux_eff
- HEV/PHEV/BEV can now account for DC:DC conversion, or extra losses for aux load on top of battery efficiency
- HEVs can apply aux load to FC and RES with diffferent efficiencies
- FCEVs when implemented can have None for the aux_eff for fuelconverters, which is more fitting than having a powertrain-level parameter that impacts FC and RES. They still will have no aux load directly supplied by the FC, only the RES, which we can check at runtime #270


kylecarow:
~TODO: Move `Default` impl into `impl_efficiency_enum!` and remove existing AuxSupplyEfficiency default impl~

~This will make all efficiencies default to constant 100% if not provided in the yaml, which will be useful if we add efficiencies to components in the future that don't currently have them~

Done

fastsim-core/src/vehicle/powertrain/fuel_converter.rs
```diff
@@ -37,6 +37,11 @@ pub struct FuelConverter {
     pub(crate) pwr_for_peak_eff: si::Power,
     /// idle fuel power to overcome internal friction (not including aux load) \[W\]
     pub pwr_idle_fuel: si::Power,
+
+    /// Efficiency to apply to aux load
+    #[serde(default)]
```
robinsteuteville, line 42: will this change the behavior for older fastsim-3 vehicle scripts? (i.e. if the vehicle script has an alt_eff but not a aux_supply_eff, will the model read in the alt_eff as the aux_supply_eff, or simply provide a default for aux_supply_eff with the user needing to update the yaml file before old runs with the file will match new runs?)

fastsim-core/src/vehicle/conv.rs
```diff
@@ -210,7 +222,29 @@ impl ConventionalVehicle {
...
+                alt_eff
+            );
+            match &self.fc.aux_supply_eff {
+                AuxSupplyEfficiency::Constant(interp) => {
+                    if interp.0 != alt_eff {
+                        // if provided both alt_eff != 1.0 and aux_eff that is not equivalent
+                        // emit warning that aux_eff overrides alt_eff
+                        eprintln!(
+                            "Warning: provided deprecated field `alt_eff` = `{}` is being overridden by `fc.aux_supply_eff`. The `alt_eff` field will be removed in a future release.",
+                            alt_eff
```
kylecarow, line 242: we need to make sure this actually sets aux_eff into aux_supply_eff (it probably already fastsim2 to 3 conversion)

kylecarow:
@robinsteuteville the issue you discovered should be fixed now!

kylecarow:
@jakeholden was hoping to bound this PR off you from a conceptual side - mind taking a look? Happy to coordinate as needed.

robinsteuteville (approved):
Looks good to me!

PR #321, opened 2026-08-12
fixing range and uf bugs

robinsteuteville:
This PR makes the following fixes to the get_label_fe() function:
1. Makes UF an optional PHEV-only return (before, it returned values for all powertrains not just PHEVs). NOTE, if there is any code that expects a UF value for other powertrain types besides PHEV, this change will break that code. However, doesn't really make sense for UF to be used in non-PHEV contexts, and would have just been returned as 0 or 1 before anyway.
2. Correctly calculates Conv and HEV range instead of those ranges being 0.
3. Updates BEV range to take into account usable battery kWh rather than total battery kWh


kylecarow (approved):
Looks good, nice work Robin! I'll go ahead & merge

Issue #326, opened 2026-08-20
Native stack overflow in SimDrive::set_ach_speed: unbounded self-recursion on infeasible steps


jhoshiko:
`sd.walk()` can kill the host process with a native stack overflow (Windows exit 0xC00000FD, SIGSEGV on Linux) with no Python exception raised. I was able to trace the stack overflow to `set_ach_speed`, which recurses on itself (fastsim-core/src/simdrive/mod.rs:735) with no depth cap, and its only exit is the `cyc_met` check, a hard `pwr_tractive <= pwr_prop_fwd_max` with no tolerance. We could add a recursion/iteration cap on the correction (mirroring `ach_speed_max_iter`), or perhaps revise the criteria for stopping the recursion. 

Reproduce:

```python
import fastsim as fs
veh = fs.Vehicle.from_file("vehicles/2016 TOYOTA Corolla 4cyl 2WD.yaml")
veh.set_save_interval(1)
cyc = fs.Cycle.from_file("cycles/<any trip id from manifest>.yaml")
fs.SimDrive(veh, cyc).walk()  # process dies natively
```

[set_ach_speed](https://github.com/NatLabRockies/fastsim/blob/7bd6ed7e0c5c2153c6ac9f7c2d10f5e246115b64/fastsim-core/src/simdrive/mod.rs#L613)

jhoshiko:
Here are some cycles, vehicles, and a manifest for reproducing the errors:

[manifest.csv](https://github.com/user-attachments/files/31279145/manifest.csv)

[vehicles.zip](https://github.com/user-attachments/files/31279147/vehicles.zip)

[cycles.zip](https://github.com/user-attachments/files/31279149/cycles.zip)

PR #335, opened 2026-09-16
Cap achieved-speed re-solves to fix native stack overflow in set_ach_speed

jhoshiko:
`set_ach_speed` recursed without bound on some real-world trips, killing the process with a stack overflow that Python could not catch. This patch adds a maximum-depth sim param and a guard that returns a catchable error instead.

- Adds `SimParams.ach_speed_max_depth` (default 100)
- Adds a depth counter through a private helper, `SimDrive::set_ach_speed_at_depth`, which the public `set_ach_speed` now wraps to avoid changes to the function signature
- The default value of 100 was chosen after a sweep across 312k real trips over 40 vehicles, with the deepest terminating trip needing 3 nested solves, and on the 337 crash cycles from the original report the deepest that still terminates needed 50. Nothing terminates between 50 and 100, and the crash cycles hit the cap at 100 and again at a probe cap of 500, so they are genuinely unbounded.
- Adds a regression test with a synthetic infeasible cycle and a cap set to 1 to test the new depth guard


jhoshiko:
Closes #326

robinsteuteville (approved):
Looks good to me!

